PR #74, opened 2026-02-16
feat(backup): add option to skip media download for some chats

Farzadd:
## Summary

For users with certain chats or groups that generate a lot of media, this option adds the ability to skip media download for those chats only

-

## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

## Database Changes

- [X] No database changes

## Data Consistency Checklist

- [X] All `chat_id` values use marked format (via `_get_marked_id()`)
- [X] All datetime values pass through `_strip_tz()` before DB operations
- [X] INSERT and UPDATE operations handle the same fields identically

## Testing

- [X] Tests pass locally (`python -m pytest tests/ -v`)
- [X] Linting passes (`ruff check .`)
- [X] Formatting passes (`ruff format --check .`)
- [X] Manually tested in development environment

## Security Checklist

- [X] No secrets or credentials committed
- [X] User input properly validated/sanitized
- [X] Authentication/authorization properly checked

## Deployment Notes

- *Includes an update to the precommit hooks configuration as previous config did not support py314 as the target version for ruff*

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,6 +51,11 @@ BATCH_SIZE=100
 # Example: SKIP_MEDIA_CHAT_IDS=-1001234567890,-1009876543210
 # SKIP_MEDIA_CHAT_IDS=
 
+# Delete existing media files and DB records for chats in SKIP_MEDIA_CHAT_IDS
+# When enabled (default), reclaims storage by removing already-downloaded media
+# Set to false to keep existing media but skip future downloads
+SKIP_MEDIA_DELETE_EXISTING=true
+
 # Hour (0-23) to recalculate backup statistics daily
 # STATS_CALCULATION_HOUR=3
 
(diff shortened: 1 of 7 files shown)


GeiserX:
Thanks for the PR! The skip logic looks solid, but there's a missing piece:

When `SKIP_MEDIA_CHAT_IDS` is enabled for a chat that **already has downloaded media**, the existing media files on disk (and their database records) are left untouched. This means you end up in an inconsistent state — old media is kept but new media is never downloaded.

**Requested change:** When a chat is in `SKIP_MEDIA_CHAT_IDS`, the backup process should also **delete existing media files from disk and remove the corresponding media records from the database** for those chats. This way, enabling the skip actually reclaims the storage those files were using, which is likely the whole reason someone would want to skip media for a chat in the first place.

This could run once at startup or at the beginning of each backup cycle — detect chats in the skip list that still have media records, delete the files, and clean up the DB rows.

GeiserX:
Merged! Thanks for the contribution @Farzadd, great feature and clean implementation. I'll push a follow-up commit to handle an edge case with media deduplication and then cut a v6.3.0 release including this.

PR #80, opened 2026-02-24
feat: multi-user viewer access control with admin UI

PhenixStar:
## Summary

- **Multi-user authentication**: Master role (env-var credentials) + DB-backed viewer accounts with per-user chat whitelists
- **Admin settings UI**: Cog icon in sidebar for master users — manage viewer accounts, assign chat access, view audit logs
- **Per-user chat filtering**: Each viewer only sees their assigned chats across all API endpoints and WebSocket
- **Audit logging**: Tracks viewer API access (endpoint, chat, IP, timestamp)
- **Security**: PBKDF2-SHA256 (600k iterations, per-user salt), timing-safe comparison, 24h session TTL
- **Backward compatible**: Existing single-user deployments work unchanged

### Backend (`src/web/main.py`)
- Dual-mode login: DB viewer accounts first, env-var master fallback
- In-memory session store with 24h TTL eviction
- `_get_user_chat_ids()` replaces `config.display_chat_ids` on all user-facing endpoints
- Admin CRUD: `GET/POST/PUT/DELETE /api/admin/viewers`, `GET /api/admin/chats`, `GET /api/admin/audit`
- CORS updated for PUT/DELETE methods

### Database (`src/db/`)
- `ViewerAccount` and `ViewerAuditLog` ORM models
- 7 new adapter methods (viewer CRUD + audit log)
- Alembic migration 007

### Frontend (`src/web/templates/index.html`)
- Settings panel: viewer management table, multi-select chat picker, activity log tab
- Logout button for all users
- Footer credit updated

### Tests (`tests/test_auth.py`)
- 28 new tests across 6 classes: password hashing, multi-user auth, admin endpoints, per-user filtering, backward compat, audit logging
- 105 total tests pass

## Test plan
- [x] All 105 tests pass (`pytest tests/ -v`)
- [x] Lint clean (`ruff check . && ruff format --check .`)
- [x] Backward compat: auth disabled = full access, single-user env-var login still works
- [x] Master respects DISPLAY_CHAT_IDS if set
- [ ] Manual test: create viewer account, verify restricted chat list
- [ ] Manual test: audit log records viewer API access

### DIFF
--- a/src/web/templates/index.html
+++ b/src/web/templates/index.html
@@ -457,7 +457,7 @@
                                         <!-- Footer -->
                                         <div class="pt-3 mt-3 border-t border-gray-700 text-center">
                                             <div class="text-[10px] opacity-70">
-                                                Made with frustration by <a href="https://github.com/PhenixStar/telegram-viewer" target="_blank" class="text-blue-400 hover:text-blue-300">phenix</a>
+                                                Made by <a href="https://github.com/GeiserX" target="_blank" class="text-blue-400 hover:text-blue-300">GeiserX</a> · Contributed by <a href="https://github.com/PhenixStar" target="_blank" class="text-blue-400 hover:text-blue-300">phenix</a>
                                             </div>
                                         </div>
                                     </div>


GeiserX:
## 🔴 Security Issues

### CRITICAL: S1 — Static `/media/` Mount Bypasses All Authentication

The `StaticFiles` mount serves all backed-up media (photos, videos, documents, voice notes) **without any auth check**:

```python
app.mount("/media", StaticFiles(directory=config.media_path), name="media")
```

FastAPI static file mounts don't participate in dependency injection, so `require_auth` is never called. Any unauthenticated user can directly access:
```
GET /media/chat_-1001234567890/photo_123.jpg
```

This also **completely bypasses viewer `allowed_chat_ids` restrictions** — a viewer restricted to Chat A can fetch media from Chat B by guessing the path.

**Fix**: Replace the blanket `StaticFiles` mount with an authenticated endpoint that checks per-user permissions before serving files. The thumbnail endpoint pattern (`dependencies=[Depends(require_auth)]`) shows the right approach.

---

### CRITICAL: S2 — Path Traversal in Thumbnail Endpoint

`folder` and `filename` in `/media/thumb/{size}/{folder:path}/{filename}` are not validated against `..` sequences. Starlette's `{folder:path}` converter accepts directory traversal:

```
GET /media/thumb/200/../../etc/hostname/something.png
```

`Path(media_root) / "../../etc"` resolves outside the media directory. While `_is_image()` restricts to image extensions, any file with `.jpg`/`.png` extension anywhere on the filesystem can be read and thumbnailed. The thumbnail is also **written** to an attacker-controlled path.

**Fix** (in `thumbnails.py`):
```python
source = (media_root / folder / filename).resolve()
if not source.is_relative_to(media_root.resolve()):
    return None

dest = _thumb_path(media_root, size, folder, filename).resolve()
if not dest.is_relative_to(media_root.resolve()):
    return None
```

---

### HIGH: S3 — Non-Constant-Time Master Token Comparison

```python
if AUTH_TOKEN and auth_cookie == AUTH_TOKEN:
```

Python's `==` for strings short-circuits on the first differing byte. Since the master token is **permanent and deterministic** (derived from credentials with a hardcoded salt), a timing attack could extract it byte-by-byte over the network.

Viewer password verification correctly uses `secrets.compare_digest`, but the master token check does not.

**Fix**:
```python
if AUTH_TOKEN and secrets.compare_digest(auth_cookie, AUTH_TOKEN):
```

---

### HIGH: S4 — Deterministic, Non-Rotating Master Token

The master `AUTH_TOKEN` is derived deterministically:
```python
AUTH_TOKEN = hashlib.pbkdf2_hmac("sha256", f"{username}:{password}".encode(), b"telegram-archive-viewer", 600_000).hex()
```

This means:
- Token **never changes** until env vars change (requires restart)
- Token is **identical** across all deployments with the same credentials
- `/api/logout` only clears viewer sessions — master token cannot be invalidated
- Anyone who obtains the token (DevTools, logs, HTTP interception) has permanent access

**Suggested fix**: Generate a random session token for the master on each login (like viewers get), store it in the session dict. This enables real logout, rotation, and revocation.

---

### HIGH: S5 — No Rate Limiting on Login

No rate limiting, account lockout, or progressive delay on `POST /api/login`. Combined with the 4-character minimum password, the viewer password space is brute-forceable in minutes.

**Suggested fix**: Add per-IP rate limiting (e.g., `slowapi` library) or per-username lockout after N failures.

---

### MEDIUM: S6 — No CSRF Protection

Admin CRUD endpoints rely solely on `SameSite=lax` cookies. While this blocks form-based CSRF, it doesn't protect against `fetch()` from an allowed CORS origin with `credentials: 'include'`. Consider adding a `X-CSRF-Token` header check.

### MEDIUM: S7 — Audit Logging Only for Viewers, Not Master

All admin actions (creating/deleting viewers, changing passwords) by the master user are **not audit-logged**. A compromised master account operates with zero forensic trail.

GeiserX:
## 🟡 Logic & Correctness Issues

### HIGH: Frontend/Backend Parameter Mismatch — Sender Search Broken

The frontend's search filter UI has a "Sender name..." text input, but sends it as `sender_id`:

```javascript
// Frontend (index.html):
if (f.sender) url += `&sender_id=${encodeURIComponent(f.sender)}`
//                      ^^^^^^^^^ sends text value like "John" as sender_id
```

The backend `get_messages` endpoint has `sender_id: int | None` — passing a text name will either return a 422 Unprocessable Entity or be silently ignored.

Meanwhile, the global search endpoint correctly uses `sender` (string match against `User.first_name`/`last_name`/`username`), which shows the right approach.

**Fix**: Change frontend to use `sender` parameter for per-chat message search, or add a string-based sender name filter parameter to the backend.

---

### MEDIUM: In-Memory Sessions — No Bounds, No Proactive Cleanup, Lost on Restart

The `_viewer_sessions` dict has several issues:

1. **No proactive cleanup** — Sessions are only evicted on access. Repeated logins create new entries without invalidating old ones, causing unbounded memory growth.
2. **Lost on restart** — All viewer sessions disappear when the process restarts, forcing all viewers to re-login.
3. **Not shared across workers** — If uvicorn runs with multiple workers, each has its own dict. A viewer authenticated on worker A is unauthenticated on worker B.

**Suggested fix**: Add a max sessions per user (evict oldest on new login), add periodic background cleanup, and consider database-backed sessions if persistence matters.

---

### MEDIUM: `is_active` Not Checked on Session Validation

`_get_current_user` checks the in-memory session but doesn't re-validate `is_active` from the database. A deactivated viewer's existing sessions remain valid for up to 24 hours.

I see that `update_viewer_account` invalidates sessions on changes — that's good. But if `is_active` is set to 0 via a direct DB update (or if the invalidation has a race condition), the deactivated user retains access.

---

### MEDIUM: Session/Cookie TTL Mismatch

Server-side sessions expire in 24h (`_SESSION_MAX_AGE = 86400`) but cookies can live up to 30 days (`AUTH_SESSION_SECONDS` from `AUTH_SESSION_DAYS`). After 24h the client holds a cookie that always returns 401, with no user-facing explanation.

---

### LOW: LIKE Wildcards Not Escaped in Search

```python
stmt = stmt.where(Message.text.ilike(f"%{query}%"))
```

Characters `%` and `_` in user input aren't escaped. Searching for `%` matches all messages, `_` matches any single character. Not a security issue (SQLAlchemy parameterizes), but a UX bug.

**Fix**:
```python
escaped = query.replace("%", r"\%").replace("_", r"\_")
stmt = stmt.where(Message.text.ilike(f"%{escaped}%", escape="\\"))
```

---

### LOW: Error Details Leaked in 500 Responses

Multiple endpoints return `detail=str(e)` which can expose internal exception messages (SQL queries, file paths, schema info) to clients:

```python
raise HTTPException(status_code=500, detail=str(e))
```

**Fix**: Return generic "Internal server error" to clients, log the full exception server-side.

PR #88, opened 2026-03-10
feat(backup): gap-fill — detect and recover skipped messages

PhenixStar:
## Summary

Adds a gap-fill feature that detects and recovers messages skipped during previous backup runs due to API errors, rate limits, or interruptions.

**Rebased onto v7.2.0** — removed all previous UI/viewer changes since you properly implemented those in v7.2.0. This PR now focuses solely on the gap-fill feature.

### What it does

- **Gap detection**: SQL `LAG()` window function finds jumps in message ID sequences exceeding a configurable threshold (default: 50 IDs)
- **Gap recovery**: Uses `iter_messages(min_id, max_id)` to fetch missing messages within detected gaps
- **Scheduler integration**: Runs automatically after each scheduled backup when `FILL_GAPS=true`
- **CLI command**: `telegram-archive fill-gaps [--chat-id ID] [--threshold N]` for manual runs

### Files changed

| File | Change |
|------|--------|
| `src/db/adapter.py` | `detect_message_gaps()` + `get_chats_with_messages()` |
| `src/telegram_backup.py` | `_fill_gap_range()`, `_fill_gaps()`, `run_fill_gaps()` |
| `src/config.py` | `FILL_GAPS` + `GAP_THRESHOLD` env vars |
| `src/scheduler.py` | Post-backup gap-fill integration |
| `src/__main__.py` | `fill-gaps` CLI subcommand |
| `tests/test_gap_fill.py` | 24 tests covering all phases |

### Configuration

```env
FILL_GAPS=true          # Enable gap-fill after each backup (default: false)
GAP_THRESHOLD=50        # Minimum gap size to investigate (default: 50)
```

### Previous review feedback

All issues from the v7.2.0-conflicting changes have been removed by rebasing. This PR contains only new, non-overlapping code.

## Test plan

- [ ] Run `telegram-archive fill-gaps` against a database with known gaps
- [ ] Enable `FILL_GAPS=true` and verify it runs after scheduled backup
- [ ] Verify gap detection query returns correct results for various gap sizes
- [ ] Run `pytest tests/test_gap_fill.py` — 24 tests pass

### DIFF
--- a/src/__main__.py
+++ b/src/__main__.py
@@ -144,6 +144,25 @@ For more information, visit: https://github.com/GeiserX/Telegram-Archive
         "--merge", action="store_true", help="Allow importing into a chat that already has messages"
     )
 
+    # Fill gaps command
+    fill_gaps_parser = subparsers.add_parser(
+        "fill-gaps",
+        help="Detect and fill message gaps from failed backups",
+        description=(
+            "Scans backed-up chats for gaps in message ID sequences "
+            "and recovers skipped messages from Telegram. "
+            "Gaps are caused by API errors, rate limits, or interruptions "
+            "during previous backup runs."
+        ),
+    )
+    fill_gaps_parser.add_argument(
+        "-c", "--chat-id", type=int, help="Fill gaps only for this specific chat ID"
+    )
+    fill_gaps_parser.add_argument(
+        "-t", "--threshold", type=int, default=None,
+        help="Minimum gap size to investigate (overrides GAP_THRESHOLD env var)",
+    )
+
     return parser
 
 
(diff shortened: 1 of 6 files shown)


GeiserX:
## 🟡 MEDIUM: Additional Functional Issues

### 1. Logout functionality removed entirely

The `performLogout` function, the logout button, and its export are all deleted. The user bar was replaced with a Settings gear icon, but Settings has no logout option. Users (both master and viewer) have **no way to log out**.

### 2. `performLogin` missing post-login init calls

After successful login, the old code ran:
```javascript
await loadFolders()
await loadArchivedCount()
setupInfiniteScroll()
```
The PR removes all three. Result: folder tabs won't appear, archived count is missing, and infinite scroll isn't initialized after login (only on page refresh).

### 3. Session cleanup task removed — memory leak

The old `session_cleanup_task()` that periodically evicted expired in-memory sessions is gone. `_viewer_sessions` only evicts entries lazily on access. Abandoned sessions persist in memory indefinitely. In a deployment with share tokens generating many sessions, this will eventually OOM.

### 4. Token verification is O(N) × 600K PBKDF2 iterations

`verify_viewer_token` loads ALL non-revoked tokens and computes PBKDF2 for each. With 50 tokens, each `/auth/token` request performs 30 million PBKDF2 iterations. An attacker can send arbitrary tokens to burn CPU. Fix: store a token prefix/hint for indexed lookup before hashing.

### 5. WebSocket "empty set = all chats" data leak

`ConnectionManager.broadcast_to_chat` has:
```python
if chat_id in subscribed_chats or not subscribed_chats:
```
An unauthenticated WebSocket connection (which the PR now accepts instead of rejecting) has an empty `subscribed_chats` set. `not set()` is `True`. These connections receive **all** broadcast messages — a data leak.

### 6. `allowed_chat_ids` return type inconsistency

`get_viewer_account_by_username` returns `allowed_chat_ids` as a raw JSON string. `get_all_viewer_accounts` returns it as a parsed list. Callers need to handle both or will get type errors.

### 7. Sponsor link and attribution removed

The "Made with heart by Sergio Fernandez" text, the `geiser.cloud` link, and the GitHub Sponsors button were removed. Replaced with "Made by GeiserX - Contributed by phenix". @GeiserX — heads up on this one, wanted to flag it for you explicitly.

### 8. `/internal/push` now allows `client_host is None`

```python
client_host in ("127.0.0.1", "localhost", "::1", None)
```
If `request.client` is `None` (certain proxy configs), the internal push endpoint becomes publicly accessible.

---

That's the complete review across all 7 areas I analyzed. Happy to discuss any of these findings or help prioritize the fixes. Again — genuinely appreciate the effort, and many of the feature ideas here are things I want in the project. The path forward is to fix the security regressions, clean up the docs, and ideally break this into smaller PRs. 🙏

PR #104, opened 2026-04-13
feat: Add proxy support for telegram client.

samnyan:
## Summary

- Allow using proxy to connect to Telegram Server.

## Type of Change

- [ ] Documentation update

## Database Changes

- [x] No database changes

## Data Consistency Checklist

## Testing

## Security Checklist

## Deployment Notes

-

## Summary by CodeRabbit

* **New Features**
  * Optional SOCKS5 proxy support for Telegram connections via environment variables.

* **Documentation**
  * Docker quick-start and Telegram credentials docs updated with proxy setup and examples.

* **Chores**
  * Added python-socks[asyncio] runtime dependency to enable proxy support.

* **Tests**
  * Added test suites covering proxy parsing, validation, and propagation into client connections.


GeiserX:
## 🔍 Comprehensive Multi-Angle Review — PR #104

**Reviewers:** Security, Architecture, Code Quality, Test Engineering, Documentation, Critical Analysis

Overall this is a well-structured contribution — clean separation, good test coverage, and thorough documentation. The issues below are all actionable and most are minor. I'd recommend addressing the **bug** and the **security** items before merge; the rest can be follow-ups.

---

### 🐛 Bug: RDNS default value mismatch across surfaces

| Surface | Default when `TELEGRAM_PROXY_RDNS` is unset |
|---|---|
| Code (`_parse_bool(proxy_rdns, default=True)`) | `True` |
| `docker-compose.yml` (`${TELEGRAM_PROXY_RDNS:-false}`) | `false` |
| `.env.example` comment | `false` |
| README reference table | `false` |

**Impact:** Users running via Docker get `rdns=false` (docker-compose substitutes before the code sees it). Users running natively without the var get `rdns=true`. This silent behavioral difference WILL confuse people debugging DNS resolution issues through their proxy.

**Fix:** Either change the code default to `False` to match all documentation, OR remove the docker-compose default (`${TELEGRAM_PROXY_RDNS:-}`) and let the code's `True` default take effect — then update all docs to say `true`.

---

### 🔒 Security Findings

**S1 — INFO-level proxy endpoint logging** *(Medium)*
```python
logger.info(
    "Telegram proxy enabled: socks5://%s:%s (rdns=%s)",
    self.telegram_proxy["addr"],
    self.telegram_proxy["port"],
    self.telegram_proxy["rdns"],
)
```
Proxy addr:port reveals internal network topology. While this project already logs phone numbers at INFO (which is its own issue), infrastructure endpoints should be at DEBUG. **CodeRabbit flagged this too — I agree it's valid.** Suggested fix:
```python
logger.info("Telegram proxy enabled (type=socks5, rdns=%s)", self.telegram_proxy["rdns"])
logger.debug("Proxy endpoint: %s:%s", self.telegram_proxy["addr"], self.telegram_proxy["port"])
```

**S2 — Port range not validated** *(Low)*
Port `0`, `-1`, `99999` all pass config validation. They'd fail later at connection time with a confusing `python-socks` error. **CodeRabbit caught this — valid.** Add after the `int()` parse:
```python
if not 1 <= parsed_port <= 65535:
    raise ValueError(f"TELEGRAM_PROXY_PORT must be between 1 and 65535, got {parsed_port}")
```

**S3 — No new credential exposure issues** *(Informational)*
Proxy username/password handling follows the same env-var pattern as `TELEGRAM_API_HASH`. Credentials aren't logged. Docker-compose passthrough is standard. No secrets committed. This is fine.

---

### 🏗️ Architecture Assessment

**A1 — Dual API surface is justified** *(No action needed)*
Module-level `build_telegram_client_kwargs()` for scripts that lack a `Config` instance, and `Config.get_telegram_client_kwargs()` for the main app. This is the right trade-off for backwards compatibility with the standalone scripts.

**A2 — Shallow copy inconsistency** *(Low)*
`Config.get_telegram_client_kwargs()` does `dict(self.telegram_proxy)` (defensive copy), but `build_telegram_client_kwargs()` returns the dict directly. If a caller mutated the returned dict, the module-level version would corrupt internal state. In practice this won't happen, but for consistency, either both should copy or neither should.

**A3 — Hard dependency + runtime check = contradiction** *(Low)*
`python-socks[asyncio]>=2.7.1` is a hard dependency (in `pyproject.toml` + `requirements.txt`), yet `build_telegram_proxy_from_env()` does a runtime `importlib.util.find_spec("python_socks")` check. If it's hard, the check is dead code. If it's meant to be optional, remove it from deps. I'd recommend: **keep it as hard dependency, remove the `find_spec` check** — the dependency is small and already required by Telethon's proxy codepath anyway.

**A4 — Forward reference changes are correct** *(No action needed)*
The `-> "TelegramConnection"` changes fix a real issue: in classmethods like `create()` and `__aenter__()`, the class name isn't available yet during class body evaluation (without `from __future__ import annotations`). These are technically bug fixes, though unrelated to proxy support. Ideally they'd be a separate commit, but not blocking.

---

### CodeRabbit Issues — Validated

| # | CodeRabbit Finding | Verdict | Reasoning |
|---|---|---|---|
| 1 | Port range validation missing | ✅ **Valid** | Real gap, should fail fast |
| 2 | Proxy logged at INFO | ✅ **Valid** | Infrastructure detail should be DEBUG |
| 3 | Tests use unittest not pytest | ❌ **Noise** | The ENTIRE existing `test_config.py` uses `unittest.TestCase`. New code correctly follows existing patterns. Refactoring to pytest is a separate task. |
| 4 | Global `sys.modules` mutation | ⚠️ **Partially valid** | `setdefault` mitigates the worst case, but a `monkeypatch` fixture would be cleaner. Low risk in practice since `src.db` import-time deps make this necessary. |
| 5 | Tests don't validate proxy dict shape | ❌ **Noise** | Tests already assert the exact dict via `assert_called_once_with(... proxy={...})`. Adding a separate shape validator is over-engineering. |

GeiserX:
## 🧪 Test Engineering Gaps

The test suite is solid for the happy path, but several edge cases and negative paths are missing:

### Missing edge case tests

| Case | Expected behavior | Why it matters |
|---|---|---|
| `TELEGRAM_PROXY_PORT=0` | Should reject (if range validation added) | Port 0 is "any available port" — not valid for a proxy target |
| `TELEGRAM_PROXY_PORT=65536` | Should reject | Out of TCP range |
| `TELEGRAM_PROXY_TYPE=SOCKS5` (uppercase) | Should work (code lowercases) | Verifies the `.lower()` actually works |
| `TELEGRAM_PROXY_TYPE=http` | Should reject with clear error | Telethon supports HTTP proxies, but this code doesn't — users should know |
| `TELEGRAM_PROXY_PASSWORD=secret` without `USERNAME` | Currently silently accepted | SOCKS5 auth requires both or neither |
| Whitespace-only `TELEGRAM_PROXY_ADDR="   "` | Currently passes validation (truthy after strip? No — `"   ".strip()` = `""` which is falsy, so it correctly fails) | Actually handled! ✅ |

### Missing no-proxy integration test

`tests/test_telegram_proxy.py` only tests the proxy-present path. There's no test verifying that `TelegramClient` is called **without** the `proxy` kwarg when proxy is not configured. This would catch regressions where an empty dict or `None` leaks through.

### `sys.modules` isolation (Low priority)

The `sys.modules.setdefault("src.db", fake_db_module)` approach works but is fragile. If test execution order changes (e.g., pytest-randomly), results could vary depending on whether `src.db` was imported first by another test file. Using `monkeypatch.setitem(sys.modules, "src.db", fake_db_module)` inside a fixture would scope the fake properly. This is a "nice to have" — not blocking.

---

## 📖 Documentation Notes

**D1 — Missing "why"**: The docs explain how to configure the proxy but not why someone would need one. A one-liner like *"Useful in regions where Telegram is blocked or when routing through a corporate network"* under the proxy section would help.

**D2 — Scope column**: All proxy vars show scope `B` (backup). But the proxy also affects the real-time listener, auth scripts, and setup. If `B` means "both" that's fine, but if it means "backup only", it's inaccurate.

**D3 — RDNS documentation says `false` default everywhere but code defaults to `True`**: Covered in the bug section of the main review above. This is the most important doc fix needed.

GeiserX (commented):
Inline suggestions from multi-angle review. See the two main review comments above for full context and CodeRabbit validation.

src/config.py
```diff
@@ -14,6 +15,84 @@
...
+        raise ValueError(f"Telegram proxy configuration is incomplete. Missing required settings: {missing}")
+
+    if proxy_type != "socks5":
+        raise ValueError("TELEGRAM_PROXY_TYPE must be 'socks5'")
+
+    try:
+        parsed_port = int(proxy_port)
+    except ValueError as e:
+        raise ValueError(f"TELEGRAM_PROXY_PORT must be a valid integer: {e}") from e
+
```
GeiserX, line 62: **Bug (Low):** Missing port range validation. Values like `0`, `-1`, or `99999` pass here and fail later with a confusing `python-socks` error.

```suggestion
    try:
        parsed_port = int(proxy_port)
    except ValueError as e:
        raise ValueError(f"TELEGRAM_PROXY_PORT must be a valid integer: {e}") from e
    if not 1 <= parsed_port <= 65535:
        raise ValueError(f"TELEGRAM_PROXY_PORT must be between 1 and 65535, got {parsed_port}")
```

src/config.py
```diff
@@ -14,6 +15,84 @@
...
+    try:
+        parsed_rdns = _parse_bool(proxy_rdns, default=True)
+    except ValueError as e:
+        raise ValueError(f"TELEGRAM_PROXY_RDNS must be a boolean value: {e}") from e
```
GeiserX, line 69: **Bug (Medium) — RDNS default mismatch:** This defaults to `True`, but `docker-compose.yml` defaults `TELEGRAM_PROXY_RDNS` to `"false"`, and `.env.example` + README both document `false` as the default.

Users running natively without the var get `rdns=True`; Docker users get `rdns=False`. Pick one — I suggest `default=False` to match all documentation:

```suggestion
        parsed_rdns = _parse_bool(proxy_rdns, default=False)
```

src/config.py
```diff
@@ -14,6 +15,84 @@
...
+
+    if importlib.util.find_spec("python_socks") is None:
+        raise ValueError(
+            "Telegram proxy support requires python-socks[asyncio]. Install it before using TELEGRAM_PROXY_* settings."
+        )
+
```
GeiserX, line 75: **Architecture (Low):** `python-socks[asyncio]` is already a hard dependency in `pyproject.toml` and `requirements.txt`, so this runtime check is dead code — `find_spec` will always succeed.

Either:
- Remove this check (since the dep is mandatory), **or**
- Move `python-socks` to an optional dependency group and keep the check

I recommend removing the check for simplicity.

src/config.py
```diff
@@ -14,6 +15,84 @@
...
+        "port": parsed_port,
+        "rdns": parsed_rdns,
+    }
+    if proxy_username:
+        proxy["username"] = proxy_username
+    if proxy_password:
+        proxy["password"] = proxy_password
+
+    return proxy
+
```
GeiserX, line 88: **Edge case (Low):** SOCKS5 auth requires both username AND password. Currently a password without username (or vice versa) is silently accepted and would fail at connection time. Consider:

```python
if proxy_username and not proxy_password:
    raise ValueError("TELEGRAM_PROXY_PASSWORD is required when TELEGRAM_PROXY_USERNAME is set")
if proxy_password and not proxy_username:
    raise ValueError("TELEGRAM_PROXY_USERNAME is required when TELEGRAM_PROXY_PASSWORD is set")
```

src/config.py
```diff
@@ -281,6 +361,13 @@ def __init__(self):
         if self.skip_media_chat_ids:
             cleanup_status = "will delete existing media" if self.skip_media_delete_existing else "keeps existing media"
             logger.info(f"Media downloads skipped for chat IDs: {self.skip_media_chat_ids} ({cleanup_status})")
+        if self.telegram_proxy:
+            logger.info(
+                "Telegram proxy enabled: socks5://%s:%s (rdns=%s)",
+                self.telegram_proxy["addr"],
+                self.telegram_proxy["port"],
+                self.telegram_proxy["rdns"],
+            )
```
GeiserX, line 372: **Security (Medium):** Proxy addr:port exposes infrastructure topology at INFO level. Move endpoint details to DEBUG; keep INFO minimal.

```suggestion
        if self.telegram_proxy:
            logger.info("Telegram proxy enabled (type=socks5, rdns=%s)", self.telegram_proxy["rdns"])
            logger.debug(
                "Proxy endpoint: %s:%s",
                self.telegram_proxy["addr"],
                self.telegram_proxy["port"],
            )
```

GeiserX:
## ✅ Second-Round Review — Post-Fix Commit `e7417f85`

All 5 inline review items from the first round have been addressed. Here's the scorecard:

| Original Finding | Status | Notes |
|---|---|---|
| **RDNS default mismatch** (Bug, Medium) | ✅ Fixed | `default=True` → `default=False`, now aligned with docker-compose and all docs |
| **Port range validation** (Bug, Low) | ✅ Fixed | Added `if not 1 <= parsed_port <= 65535` check |
| **Proxy endpoint at INFO** (Security, Medium) | ✅ Fixed | Endpoint moved to `logger.debug`, INFO now only shows `type=socks5, rdns=...` |
| **Dead `find_spec` check** (Architecture, Low) | ✅ Fixed | Removed `importlib.util` import and runtime check entirely |
| **Username/password pairing** (Edge case, Low) | ✅ Fixed | Added `bool(proxy_username) != bool(proxy_password)` XOR validation |

### New tests added — good coverage expansion

- `test_proxy_rejects_port_zero` — port 0 boundary
- `test_proxy_rejects_port_above_range` — port 65536 boundary
- `test_proxy_type_is_case_insensitive` — verifies `.lower()` works
- `test_proxy_rejects_non_socks5_type` — HTTP proxy type rejection
- `test_proxy_rejects_partial_auth_credentials` — password without username
- `test_connection_omits_proxy_when_not_configured` — no-proxy integration path

---

### Remaining minor items (non-blocking)

These are all low-priority and can be addressed in follow-ups:

1. **`sys.modules` global mutation** in `tests/test_telegram_proxy.py:12-17` — still uses module-level `sys.modules.setdefault()` instead of a scoped `monkeypatch` fixture. Works fine in practice due to `setdefault` semantics, but could cause test ordering issues with `pytest-randomly`. Low risk.

2. **Partial auth test only covers one direction** — `test_proxy_rejects_partial_auth_credentials` tests password-without-username but not username-without-password. The code (`bool(x) != bool(y)`) handles both, but only one direction is tested.

3. **Shallow copy inconsistency** — `Config.get_telegram_client_kwargs()` does `dict(self.telegram_proxy)` (defensive copy) but `build_telegram_client_kwargs()` returns the dict directly. No practical impact since callers don't mutate the result.

4. **Minor docs suggestion** — A one-liner explaining *why* someone might need a proxy (e.g., "Useful in regions where Telegram is blocked or behind corporate firewalls") would help discoverability in the README.

---

### Verdict

The fix commit cleanly addresses all substantive feedback. The remaining items are cosmetic/low-risk. This PR is in good shape for merge. Nice work @samnyan!

PR #122, opened 2026-04-18
fix: show General forum topic messages (coalesce NULL reply_to_top_id)

tondeaf:
## Summary

- Apply the same `coalesce(reply_to_top_id, 1) == topic_id` predicate in `get_messages_paginated()` that `get_forum_topics()` already uses for counting, so the General topic pane stops returning empty when its messages have NULL `reply_to_top_id` (pre-v6.2.0 / pre-forum) or explicit `reply_to_top_id=1` (post-forum-enable).
- Three regression tests covering NULL+explicit-1 General messages, strict filtering for non-General topics, and unfiltered no-topic queries.

## Type of Change

- [ ] Documentation update

## Database Changes

- [x] No database changes

## Data Consistency Checklist

- [ ] All `chat_id` values use marked format (via `_get_marked_id()`) — N/A, no chat_id construction
- [ ] All datetime values pass through `_strip_tz()` before DB operations — N/A, no datetime handling
- [ ] INSERT and UPDATE operations handle the same fields identically — N/A, change is read-side only

## Testing

## Security Checklist

- [x] User input properly validated/sanitized — `topic_id` flows through SQLAlchemy bound parameters
- [x] Authentication/authorization properly checked — same auth path as other paginated reads

## Deployment Notes

- No image rebuild needed beyond a normal release; the change is a query predicate and takes effect immediately on rolled code.

---

## The bug

In the web viewer, a forum chat's topic list correctly shows the **General** topic with its true message count (e.g. "67 messages"), but clicking the General tab renders an empty message pane.

Reproduce:
1. Enable a forum on a group chat that has pre-existing (pre-forum) messages, or keep using a forum chat whose early messages predate v6.2.0.
2. Run the archiver + viewer.
3. Open the chat → **General** topic.

Sidebar shows the correct count; main pane is empty.

## Root cause

`get_forum_topics()` in `src/db/adapter.py` already handles this correctly when *counting* General messages:

```python
effective_topic_id = func.coalesce(Message.reply_to_top_id, 1).label("effective_topic_id")
```

…with a comment that explains: *"Messages with `reply_to_top_id=NULL` are treated as General topic (id=1), since pre-v6.2.0 messages and pre-forum messages lack topic assignment and Telegram's client displays them under General."*

But `get_messages_paginated()` uses a strict equality filter:

```python
if topic_id is not None:
    stmt = stmt.where(Message.reply_to_top_id == topic_id)
```

So `?topic_id=1` never matches the NULL-top rows, and the General pane is empty despite the sidebar claim.

## The fix

Apply the same `coalesce(reply_to_top_id, 1) == topic_id` filter in `get_messages_paginated()`. General sees its NULL-top messages; non-General topics still filter strictly.

One-line change + three regression tests (`tests/test_general_topic_messages.py`) covering:

- `topic_id=1` returns NULL-top **and** explicit `reply_to_top_id=1` messages (post-review fix-up)
- non-General topic IDs still filter strictly
- no `topic_id` returns everything regardless of `reply_to_top_id`

Tests use `Base.metadata.create_all` on an in-memory SQLite engine so column drift can't silently break them.

## Verification against a real DB

On the affected forum chat (pre-forum + forum messages mixed):
- `GET /api/chats/<id>/messages?topic_id=1` — before: `[]`; after: returns the General messages.
- `GET /api/chats/<id>/messages?topic_id=<other>` — still returns only rows matching that exact topic.

## Test plan
- [x] `pytest tests/test_general_topic_messages.py` — 3/3 pass
- [x] Manually verified General tab now renders messages in the viewer
- [x] Manually verified non-General topics still filter strictly

### DIFF
--- a/src/db/adapter.py
+++ b/src/db/adapter.py
@@ -1098,9 +1098,10 @@ class DatabaseAdapter:
                 .where(Message.chat_id == chat_id)
             )
 
-            # v6.2.0: Filter by forum topic
+            # v6.2.0: Filter by forum topic. NULL reply_to_top_id == General (id=1),
+            # matching the coalesce in get_forum_topics counts.
             if topic_id is not None:
-                stmt = stmt.where(Message.reply_to_top_id == topic_id)
+                stmt = stmt.where(func.coalesce(Message.reply_to_top_id, 1) == topic_id)
 
             if search:
                 escaped = search.replace("\\", "\\\\").replace("%", r"\%").replace("_", r"\_")
--- a/tests/test_general_topic_messages.py
+++ b/tests/test_general_topic_messages.py
@@ -53,6 +53,9 @@ async def _make_adapter_with_messages():
             (3, datetime(2024, 3, 1, 10), "sexy topic msg", 47),
             (4, datetime(2024, 3, 2, 10), "another sexy msg", 47),
             (5, datetime(2024, 4, 1, 10), "photos topic", 144),
+            # Post-forum-enable: Telegram sets reply_to_top_id=1 explicitly on
+            # new General messages. Must appear alongside NULL rows under topic_id=1.
+            (6, datetime(2024, 5, 1, 10), "post-forum general (explicit 1)", 1),
         ]
         for mid, dt, body, top in rows:
             session.add(Message(id=mid, chat_id=chat_id, date=dt, text=body, reply_to_top_id=top))
@@ -62,13 +65,14 @@ async def _make_adapter_with_messages():
 
 
 @pytest.mark.asyncio
-async def test_general_topic_includes_null_reply_to_top_id():
-    """topic_id=1 (General) must match messages whose reply_to_top_id IS NULL."""
+async def test_general_topic_includes_null_and_explicit_one():
+    """topic_id=1 (General) must match both NULL reply_to_top_id (pre-forum) and
+    explicit reply_to_top_id=1 (post-forum-enable) messages."""
     adapter, engine, chat_id = await _make_adapter_with_messages()
     try:
         messages = await adapter.get_messages_paginated(chat_id=chat_id, topic_id=1)
         ids = sorted(m["id"] for m in messages)
-        assert ids == [1, 2], f"General (topic_id=1) must return the 2 NULL-top messages, got {ids}"
+        assert ids == [1, 2, 6], f"General (topic_id=1) must return NULL and explicit-1 messages, got {ids}"
     finally:
         await engine.dispose()
 
@@ -93,6 +97,6 @@ async def test_no_topic_filter_returns_all():
     adapter, engine, chat_id = await _make_adapter_with_messages()
     try:
         all_msgs = await adapter.get_messages_paginated(chat_id=chat_id)
-        assert sorted(m["id"] for m in all_msgs) == [1, 2, 3, 4, 5]
+        assert sorted(m["id"] for m in all_msgs) == [1, 2, 3, 4, 5, 6]
     finally:
         await engine.dispose()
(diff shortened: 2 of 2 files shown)


GeiserX:
Hey @tondeaf, thanks for this excellent bug fix! The root cause analysis in the PR description is really well done — identifying the mismatch between `get_forum_topics()` (coalesce) and `get_messages_paginated()` (strict equality) is spot on.

I ran a thorough review (code quality, security, test adequacy) and the fix looks great overall. A few items to address before merge:

### Code Quality (1 low-severity note)

**LOW — `coalesce()` may prevent index usage on `idx_messages_topic`**
`COALESCE(reply_to_top_id, 1) = ?` wraps the column in a function, which prevents the B-tree index from being used directly. Pragmatically this is unlikely to matter (the `chat_id` predicate narrows first, and `LIMIT 50 ORDER BY date DESC` stops early), but worth noting. An alternative that preserves index usage for non-General topics:
```python
if topic_id == 1:
    stmt = stmt.where(or_(Message.reply_to_top_id == 1, Message.reply_to_top_id.is_(None)))
else:
    stmt = stmt.where(Message.reply_to_top_id == topic_id)
```
That said, the current approach is consistent with `get_forum_topics()` which already uses the same coalesce pattern, so I'd understand keeping it as-is for consistency. Your call.

### Test Adequacy (1 missing edge case)

**Please add a test for explicit `reply_to_top_id=1`**: The current seed data only has `NULL` for General messages, but in production, after forums are enabled, Telegram sets `reply_to_top_id=1` on new General messages. Add a seed row like:
```python
(6, datetime(2024, 5, 1, 10), "explicit general msg", 1),
```
Then update `test_general_topic_includes_null_reply_to_top_id` to assert `ids == [1, 2, 6]` — proving both NULL and explicit-1 messages appear together under `topic_id=1`.

### CodeRabbit

CodeRabbit flagged that the PR description doesn't follow the repo's PR template (Type of Change checkboxes, Database Changes section, etc.). Would be nice to restructure the description to match the template format.

### Security

No concerns — the change is a parameterized SQLAlchemy expression, `chat_id` filter is unconditional, and auth is enforced before the query runs. Clean.

Thanks again for the contribution! 🙏

tondeaf:
Addressed your feedback in the latest commit. Summary of what changed and why:

**Test row for explicit `reply_to_top_id=1`**
- Added a row with `reply_to_top_id=1` (post-forum-enable General) and updated `test_general_topic_includes_null_and_explicit_one` (renamed) to expect `[1, 2, 6]`. `test_no_topic_filter_returns_all` updated to `[1, 2, 3, 4, 5, 6]`.

**Index-usage alternative (LOW)**
- Left as-is. As you noted, the current approach matches `get_forum_topics()` and the chat_id predicate + LIMIT make the practical impact negligible. Happy to switch to the `or_(eq, is_null)` form if you'd rather have it.

**PR description template**
- Not yet — wanted to keep this commit narrowly scoped. I'll restructure the body if you'd like before merge.

Ready for another look.

GeiserX:
Thanks @tondeaf! This fix is now live in [v7.6.4](https://github.com/GeiserX/Telegram-Archive/releases/tag/v7.6.4). 🎉

The coalesce alignment between `get_forum_topics()` and `get_messages_paginated()` was a great catch — General topic users were seeing a message count in the sidebar but getting zero results when clicking in.

I also shipped a follow-up in #132 that built on your test suite: renamed the test data to neutral names, converted to a pytest fixture, removed redundant decorators, and added edge case tests for nonexistent topics, `topic_id=0`, and topic+search interaction. Your original 3 tests were solid foundations for all of that.

Issue #1, opened 2025-12-07
[FEATURE] Date picker in conversation view


calvinturbo:
Would it be possible to add a date picker so you can click the date (same as in Telegram app itself), select the date and will scroll up there?

Please also post a donation link on your readme.md, I would be willing to make a small donation as I greatly appreciate your work :)

PR #123, opened 2026-04-18
fix: use bound parameters in pg_notify to survive dollar-tokens in payloads

tondeaf:
## Summary

- Switch `_notify_postgres` from f-string-built SQL to `pg_notify(:channel, :payload)` with SQLAlchemy bound parameters, so message bodies that happen to contain `$N` / `$<letter>` tokens stop tripping asyncpg's positional-placeholder parser and breaking realtime fanout.
- Regression tests consolidated into `TestRealtimeNotifierPostgres` covering payload-not-in-SQL, dollar-token end-to-end (no warning), single-quote payload (replaces the old manual `.replace("'", "''")` escaping), and a tightened pg_notify-shape assertion to replace the stale `execute-was-called` test.

## Type of Change

- [ ] Documentation update

## Database Changes

- [x] No database changes

## Data Consistency Checklist

- [ ] All `chat_id` values use marked format (via `_get_marked_id()`) — N/A, no chat_id construction in this path
- [ ] All datetime values pass through `_strip_tz()` before DB operations — N/A, no datetime handling in this path
- [ ] INSERT and UPDATE operations handle the same fields identically — N/A, change is in NOTIFY plumbing, not in row writes

## Testing

## Security Checklist

- [x] User input properly validated/sanitized — payload now flows as a bound parameter, eliminating the manual single-quote escaping path entirely
- [x] Authentication/authorization properly checked — change is below the auth boundary; no surface change

## Deployment Notes

- Cuts over on the next service restart; no migration. The fix is purely safer SQL plumbing for an existing call.

---

## The bug

Realtime LISTEN/NOTIFY fanout to the viewer silently breaks for any chat when a single message body contains a `$N` or `$<letter>` token (e.g. `"sshhhh $1 will take LONGER!"` or `"paid $D today"`). The viewer stops receiving updates for that chat until the service is restarted. The only visible symptom is a vague "Failed to send realtime notification" warning in the log.

## Root cause

`_notify_postgres` in `src/realtime.py` f-string-interpolates the JSON payload directly into the SQL string passed to `text()`:

```python
sql = f"NOTIFY telegram_updates, '{json.dumps(payload)}'"
await session.execute(text(sql))
```

asyncpg scans the raw SQL for `$N` positional placeholders **before** dispatching to Postgres, so any message text containing tokens that look like parameters trips its parser:

```
asyncpg.exceptions.PostgresSyntaxError: there is no parameter $1
```

## The fix

Use `pg_notify(:channel, :payload)` with SQLAlchemy bound parameters. Payload becomes opaque string data to the driver — no SQL-level quoting, and the placeholder scanner never sees the payload content.

```python
stmt = text("SELECT pg_notify(:channel, :payload)")
await session.execute(stmt, {"channel": channel, "payload": payload})
```

Regression tests live alongside the existing `TestRealtimeNotifierPostgres` class:

- Payload containing `$1`/`$D` tokens is never embedded in the final SQL.
- End-to-end: `.notify()` with those tokens does not emit the "Failed to send realtime notification" warning.
- Single-quote payload round-trips cleanly (covers the removal of the old `.replace("'", "''")` path).
- The pg_notify SQL shape and `channel`/`payload` bound-param keys are explicitly asserted (replacing the stale `execute-was-called` test).

## Test plan
- [x] `pytest tests/test_realtime.py::TestRealtimeNotifierPostgres` — 5/5 pass; full `tests/test_realtime.py` 52/52 pass
- [x] Manually verified in production: the same message body that previously killed realtime updates now round-trips cleanly to the viewer.


GeiserX:
Hey @tondeaf, great catch on this one! The `$N` placeholder parsing by asyncpg is a really subtle bug — using `pg_notify` with bound parameters is exactly the right fix. The expanded docstring explaining the *why* is appreciated too.

Here's the review:

### Security (APPROVE)

This is actually a security improvement. The old code did `text(f"NOTIFY telegram_updates, '{payload_json}'")` with manual `.replace("'", "''")` — while the single-quote escaping was technically correct for PostgreSQL string literals, it was fragile and had the asyncpg `$N` parsing issue. The new `pg_notify(:channel, :payload)` with bound params eliminates the entire class of problems. No injection surface remains.

### Code Quality (1 HIGH, 1 MEDIUM, 1 LOW)

**HIGH — Existing test `test_notify_postgres_executes_notify_command` has a stale docstring and doesn't validate the new contract**
The existing test at `tests/test_realtime.py:283` says `"""_notify_postgres sends NOTIFY via SQL."""` — but after this PR, it sends `SELECT pg_notify(...)`. The test only asserts `execute.assert_called_once()` and `commit.assert_called_once()`, not the SQL structure. Please either:
- Update the docstring and add an assertion that the first arg contains `pg_notify` and the second arg is a dict with `channel`/`payload` keys, or
- Merge the new tests into the existing `TestRealtimeNotifierPostgres` class.

**MEDIUM — Duplicate mock setup**
`_make_notifier_with_fake_session()` in the new file duplicates the setup already in `tests/test_realtime.py:285-295`. Consider extracting a shared fixture or merging the new tests into the existing class.

**LOW — Opaque ticket ID in docstring**
The test file references `(yrru-ks3d)` which is meaningless to outside contributors. Replace with the PR/issue number.

### Test Adequacy (2 must-fix, 2 should-fix)

**Must-fix:**
1. **Add `commit()` assertion** to `test_notify_does_not_interpolate_payload_into_sql`. The `session.commit()` call is critical (NOTIFY fires on commit in PostgreSQL) but neither test verifies it:
   ```python
   assert session.commit.await_count == 1
   ```

2. **Remove `@pytest.mark.asyncio` decorators** — the project uses `asyncio_mode = "auto"` in pyproject.toml, so they're redundant. Minor but keeps consistency.

**Should-fix:**
3. **Add a single-quote payload test** — the old code explicitly handled single quotes with `.replace("'", "''")`. The new code drops that escaping. A test proving single quotes work with bound params documents this intentional change.

4. **Simplify mock setup** — remove the explicit `__aenter__`/`__aexit__` assignments (lines 21-22). `AsyncMock` already provides them.

### CodeRabbit

CodeRabbit was rate-limited on this PR. I'll retrigger it once the cooldown expires.

Thanks again for the contribution! 🙏

GeiserX:
@coderabbitai full review

tondeaf:
Addressed your feedback in the latest commit. Summary:

**HIGH — stale `test_notify_postgres_executes_notify_command`**
- Replaced with `test_notify_postgres_uses_pg_notify_with_bound_params`, which validates the new contract: SQL contains `pg_notify`, params dict has `channel`/`payload` keys.

**MEDIUM — duplicate mock setup**
- Merged everything from `tests/test_realtime_notify.py` into the existing `TestRealtimeNotifierPostgres` class via a static `_make_notifier_with_fake_session` helper. Deleted the standalone file.

**LOW — opaque ticket id**
- `(yrru-ks3d)` removed; docstring now references PR #123.

**Must-fix #1 — `commit()` assertion**
- Added `session.commit.await_count == 1` to the dollar-token regression test (and to the new pg_notify shape test).

**Must-fix #2 — `@pytest.mark.asyncio` decorators**
- Dropped (project uses `asyncio_mode = "auto"`).

**Should-fix #3 — single-quote payload test**
- Added `test_notify_handles_single_quotes_in_payload`.

**Should-fix #4 — explicit `__aenter__`/`__aexit__` setup**
- One small pushback: dropping the line entirely breaks the tests. `AsyncMock()` provides the protocol, but `__aenter__()` returns a fresh child mock by default rather than `self`, so `async with session as s:` binds `s` to a different mock and the assertions miss the real call. I kept `session.__aenter__.return_value = session` and documented why in the helper docstring. The `__aexit__` setup is genuinely unneeded and was removed.

Ready for another look.

GeiserX:
Thanks for the contribution, @tondeaf! Merged and shipped a follow-up in #131 (v7.6.3) addressing edge cases found during review:

- **Edit notifications now truncated too** — your fix exposed a pre-existing gap: `data["new_text"]` (edit path) bypassed the 500-char truncation, so a 4096-char emoji edit could produce a 16KB payload exceeding PostgreSQL's 8KB NOTIFY limit. Now handled via shared `_truncate_notify_data()`.
- **Timing-safe push secret comparison** — `/internal/push` bearer token check switched from `!=` to `secrets.compare_digest()`.
- **Stable test assertions** — `str(stmt)` → `stmt.text` for SQLAlchemy `TextClause`.

Released as [v7.6.3](https://github.com/GeiserX/Telegram-Archive/releases/tag/v7.6.3).

PR #118, opened 2026-04-17
feat: exclude specific topics in forum supergroups

GeiserX:
## Summary

- Adds `SKIP_TOPIC_IDS` env var to exclude specific forum topics from backup without excluding the entire chat
- Format: `chat_id:topic_id,chat_id:topic_id,...` (e.g. `-1001234567890:42,-1001234567890:1337`)
- Filtering applied consistently in scheduled backup, gap-fill, forum topic metadata, and real-time listener (new messages + edits)
- Deletion handler intentionally excluded since `MessageDeleted` events lack topic info

## Changes

- **`src/config.py`**: New `_parse_topic_skip_list()` parser, `skip_topic_ids` property, `should_skip_topic()` method, and startup logging
- **`src/telegram_backup.py`**: Topic-level filtering in `_backup_dialog`, `_fill_gap_range`, and `_backup_forum_topics` (both API and fallback paths)
- **`src/listener.py`**: Topic-level filtering in `on_new_message` and `on_message_edited` handlers, plus startup logging
- **`tests/test_config.py`**: 12 new tests covering parsing, validation, and `should_skip_topic()` behavior
- **`README.md`**, **`docker-compose.yml`**, **`.env.example`**: Documentation for the new env var

## Test plan

- [x] All 49 config tests pass (`pytest tests/test_config.py -v`)
- [ ] Manual test with a forum supergroup containing multiple topics
- [ ] Verify excluded topic messages are skipped in scheduled backup
- [ ] Verify excluded topic messages are skipped in real-time listener
- [ ] Verify non-excluded topics in the same chat continue to be backed up

Closes #117

## Release Notes

* **New Features**
  * Added support for excluding specific forum topics from backups via the `SKIP_TOPIC_IDS` environment variable (format: `chat_id:topic_id,...`). Applies to both scheduled backups and real-time updates.

* **Documentation**
  * Updated configuration documentation and Docker Compose template to include topic filtering options.

PR #124, opened 2026-04-18
feat: make Telethon flood-waits visible in the log

tondeaf:
## Summary

- Set `flood_sleep_threshold=0` on the shared `TelegramClient` so every `FloodWaitError` surfaces, then wrap `client.iter_messages` at the two bulk call-sites in `src/telegram_backup.py` with `iter_messages_with_flood_retry`, which logs the wait, sleeps, and resumes from the last yielded id.
- Bound the wrapper: `MAX_FLOOD_RETRIES=5` consecutive flood-waits without progress (counter resets on each yield) before the wrapper gives up and re-raises; `MAX_FLOOD_WAIT_SECONDS=600` clamp on `e.seconds` to defend against pathological values.
- New `FLOOD_WAIT_LOG_THRESHOLD` env var (default 10s) suppresses noisy short-wait log lines while keeping the operator-visible long ones.

## Type of Change

- [ ] Documentation update

## Database Changes

- [x] No database changes

## Data Consistency Checklist

## Testing

## Security Checklist

- [x] User input properly validated/sanitized — `e.seconds` is integer from Telethon and is clamped before use
- [x] Authentication/authorization properly checked — change is internal to the backup pipeline; no auth surface modified

## Deployment Notes

- New optional env var: `FLOOD_WAIT_LOG_THRESHOLD` (default `10`, set to `0` to log every wait). No-op for existing deployments — the defaults match the prior log-everything behavior except for sub-10-second waits which are now silent.

---

## The problem

Telethon's `TelegramClient` silently sleeps through any `FloodWaitError` whose wait is ≤ `flood_sleep_threshold` (default 60 s). On a large `iter_messages` pass — e.g. initial backup of a 100 k-message chat — Telegram often hands back much longer waits, and the operator can't tell "crashed" from "rate-limited":

- No log lines
- No DB writes
- Realtime listener is still firing
- `systemctl status` still shows the service as active

## The fix

1. Set `flood_sleep_threshold=0` on the shared `TelegramClient` so every `FloodWaitError` bubbles up as an exception rather than a silent sleep.
2. Add `iter_messages_with_flood_retry(client, entity, **kwargs)` — a thin async generator that wraps `client.iter_messages`, logs `FloodWait: sleeping Ns before resuming (last_msg_id=M, retry=N/MAX)`, sleeps, then resumes iteration from the last-seen message id.
3. Bound the wrapper after the original review: `MAX_FLOOD_RETRIES=5` consecutive flood-waits without progress before re-raising (the counter resets every time iteration yields a message, so a long backfill that hits one wait per chunk doesn't trip the cap), and `MAX_FLOOD_WAIT_SECONDS=600` clamp on the sleep so a buggy/malicious server can't hang the process for hours.
4. Use that wrapper at the two bulk-iteration call-sites in `src/telegram_backup.py` (full chat backup and gap-fill).

`FLOOD_WAIT_LOG_THRESHOLD` (default 10 s) controls the minimum wait to log, so trivial 1-2 s floods don't spam the log.

Before/after is qualitative:

- **Before:** multi-minute flood-wait looks like a hang.
- **After:** `FloodWait: sleeping 47s before resuming (last_msg_id=87342, retry=1/5)` — operator immediately knows what's happening.

## Conflict notes

Rebased on top of the recently-merged `SKIP_TOPIC_IDS` feature (#118). Both conflict sites in `src/telegram_backup.py` are trivial — our flood-retry wrapper wraps the `iter_messages` call, and the topic-skip check stays inside the loop body where upstream added it.

## Test plan
- [x] `pytest tests/test_flood_wait_visibility.py` — 10/10 pass (3 original + 7 added in review)
- [x] `pytest tests/test_config.py` — passes
- [x] Manually triggered in production on a large chat; log now shows the flood-wait line with `retry=N/MAX` and resumes cleanly from the correct message id.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -89,10 +89,11 @@ def build_telegram_proxy_from_env() -> dict | None:
 
 def build_telegram_client_kwargs() -> dict:
     """Build common Telethon client keyword arguments from environment configuration."""
+    kwargs: dict = {"flood_sleep_threshold": 0}
     proxy = build_telegram_proxy_from_env()
-    if proxy is None:
-        return {}
-    return {"proxy": dict(proxy)}
+    if proxy is not None:
+        kwargs["proxy"] = dict(proxy)
+    return kwargs
 
 
 class Config:
@@ -602,10 +603,16 @@ class Config:
             )
 
     def get_telegram_client_kwargs(self) -> dict:
-        """Get shared TelegramClient keyword arguments."""
-        if self.telegram_proxy is None:
-            return {}
-        return {"proxy": dict(self.telegram_proxy)}
+        """Get shared TelegramClient keyword arguments.
+
+        ``flood_sleep_threshold=0`` forces Telethon to raise FloodWaitError
+        instead of silently sleeping, so long waits become visible in the log
+        via ``iter_messages_with_flood_retry``.
+        """
+        kwargs: dict = {"flood_sleep_threshold": 0}
+        if self.telegram_proxy is not None:
+            kwargs["proxy"] = dict(self.telegram_proxy)
+        return kwargs
 
 
 def setup_logging(config: Config):
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -38,28 +38,60 @@ from .message_utils import extract_topic_id
 logger = logging.getLogger(__name__)
 
 
+MAX_FLOOD_RETRIES = 5
+MAX_FLOOD_WAIT_SECONDS = 600
+
+
 async def iter_messages_with_flood_retry(client, entity, *, min_id=0, **kwargs):
     """Wrap ``client.iter_messages`` so FloodWaitError is logged and retried.
 
     With ``flood_sleep_threshold=0`` on the client, every flood-wait bubbles up
     as an exception. We log the wait and resume iteration from the last yielded
     message id so progress isn't lost.
+
+    Bounded retries: the inner ``while`` is capped at ``MAX_FLOOD_RETRIES``
+    *consecutive* flood-waits without progress, and the counter resets every
+    time iteration yields a message. Without the cap, an account-restricted
+    Telegram session would loop forever on one chat and block every later one.
+
+    Bounded sleep: ``e.seconds`` is clamped at ``MAX_FLOOD_WAIT_SECONDS`` so a
+    pathologically large value (buggy or malicious server response) cannot
+    hang the process for hours.
+
+    The ``FLOOD_WAIT_LOG_THRESHOLD`` env var (default 10) suppresses log
+    output for short waits — those are routine and noisy in healthy backfills.
+    Set to 0 to log every wait.
     """
+    log_threshold_seconds = int(os.getenv("FLOOD_WAIT_LOG_THRESHOLD", "10"))
     resume_from = min_id
+    retries = 0
     while True:
         try:
             async for msg in client.iter_messages(entity, min_id=resume_from, **kwargs):
                 yield msg
                 if getattr(msg, "id", None) is not None:
                     resume_from = max(resume_from, msg.id)
+                retries = 0
             return
         except FloodWaitError as e:
-            logger.warning(
-                "FloodWait: sleeping %ss before resuming (last_msg_id=%s)",
-                e.seconds,
-                resume_from,
-            )
-            await asyncio.sleep(e.seconds + 1)
+            retries += 1
+            if retries > MAX_FLOOD_RETRIES:
+                logger.error(
+                    "FloodWait: exceeded %d retries without progress, giving up (last_msg_id=%s)",
+                    MAX_FLOOD_RETRIES,
+                    resume_from,
+                )
+                raise
+            wait_seconds = min(e.seconds, MAX_FLOOD_WAIT_SECONDS)
+            if e.seconds >= log_threshold_seconds:
+                logger.warning(
+                    "FloodWait: sleeping %ss before resuming (last_msg_id=%s, retry=%d/%d)",
+                    wait_seconds,
+                    resume_from,
+                    retries,
+                    MAX_FLOOD_RETRIES,
+                )
+            await asyncio.sleep(wait_seconds + 1)
 
 
 class TelegramBackup:
@@ -695,7 +727,6 @@ class TelegramBackup:
             if self.config.should_skip_topic(chat_id, extract_topic_id(message)):
                 continue
 
-
             msg_data = await self._process_message(message, chat_id)
             batch_data.append(msg_data)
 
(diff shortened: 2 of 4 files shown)


GeiserX:
Hey @tondeaf, excellent feature! Making flood-waits visible is a huge UX improvement for operators — "is it crashed or rate-limited?" is a really common pain point with Telethon. The `min_id` extraction via keyword-only parameter (`*, min_id=0, **kwargs`) is a smart design choice that prevents double-passing.

Here's the thorough review:

### Code Quality (1 CRITICAL, 2 MEDIUM, 2 LOW)

**CRITICAL — `while True` loop has no retry cap**
If Telegram repeatedly returns `FloodWaitError` (e.g., account temporarily restricted), the loop retries forever. The backup process hangs indefinitely on a single chat, blocking all subsequent chats. Please add a maximum retry count:
```python
MAX_FLOOD_RETRIES = 5
retries = 0
while True:
    try:
        async for msg in client.iter_messages(entity, min_id=resume_from, **kwargs):
            yield msg
            if getattr(msg, "id", None) is not None:
                resume_from = max(resume_from, msg.id)
            retries = 0  # reset on successful progress
        return
    except FloodWaitError as e:
        retries += 1
        if retries > MAX_FLOOD_RETRIES:
            logger.error("FloodWait: exceeded %d retries, giving up (last_msg_id=%s)", MAX_FLOOD_RETRIES, resume_from)
            raise
        logger.warning(...)
        await asyncio.sleep(e.seconds + 1)
```

**MEDIUM — PR body advertises `FLOOD_WAIT_LOG_THRESHOLD` but it's not implemented**
The description says "New config knob `FLOOD_WAIT_LOG_THRESHOLD` (default 10s) controls the minimum wait to log" — but this doesn't exist in the diff. All flood-waits are logged unconditionally. Either implement it or remove the claim from the PR body.

**MEDIUM — `e.seconds` is trusted without an upper bound**
A malicious/buggy server could send an extremely large `FloodWaitError.seconds` value, hanging the process for hours/days. Cap the sleep:
```python
MAX_FLOOD_WAIT_SECONDS = 600  # 10 minutes
wait_seconds = min(e.seconds + 1, MAX_FLOOD_WAIT_SECONDS)
await asyncio.sleep(wait_seconds)
```

**LOW — Two extra blank lines introduced**
At `telegram_backup.py:698` and `:790`, there are two consecutive blank lines after the topic-skip blocks. PEP 8 recommends at most one within a function body.

**LOW — `(yrru-mix3)` in test docstring**
Same as PR #123 — opaque internal ID, replace with PR/issue number.

### Security (APPROVE with above changes)

- No rate-limit amplification — the wrapper sleeps the full requested duration before retrying.
- Log line only exposes `e.seconds` (int) and `resume_from` (message ID int) — no PII, no entity names.
- `flood_sleep_threshold=0` is the right call for a backup tool where visibility matters.

### Test Adequacy (2 HIGH gaps, 2 MEDIUM)

**HIGH — No test for FloodWait before any message is yielded**
The `seeded_iter` always yields `id=1` before the error. But what if the very first API call raises FloodWaitError? `resume_from` stays at original `min_id`. The initial `fake_iter` was defined for this but is actually dead code (overwritten by `seeded_iter` before use). Please: remove dead `fake_iter`, add a separate test for the no-yield-before-flood case.

**HIGH — No test for multiple consecutive FloodWaitErrors**
Production commonly sees 3-4 consecutive floods before success. Add a test that raises FloodWaitError 2-3 times, then succeeds.

**MEDIUM — No test for `max_id` preservation across retries**
The gap-fill call-site passes `max_id=gap_end` via `**kwargs`. It's forwarded correctly, but an explicit test would catch regressions.

**MEDIUM — `_fake_db` fixture is overly broad**
The `autouse=True` fixture reloads `src.connection` and `src.telegram_backup` for every test, including `test_config_kwargs_include_flood_sleep_threshold_zero` which doesn't need it. Consider narrowing the scope.

### CodeRabbit

CodeRabbit was rate-limited on this PR. I'll retrigger it once the cooldown expires.

Thanks for the contribution! 🙏

GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

tondeaf:
Addressed your feedback in the latest commit. Summary:

**CRITICAL — unbounded retry loop**
- Added `MAX_FLOOD_RETRIES = 5` with a counter that resets on every successful yield. A long backfill that hits one flood-wait per chunk doesn't trip the cap; only consecutive flood-waits without progress count, and after the cap the wrapper logs and re-raises so the chat fails closed instead of hanging.

**MEDIUM — `e.seconds` upper bound**
- Added `MAX_FLOOD_WAIT_SECONDS = 600`; `e.seconds` is clamped before we sleep.

**MEDIUM — `FLOOD_WAIT_LOG_THRESHOLD` advertised but not implemented**
- Implemented. Reads the env var (default 10s); waits below the threshold are silent, the rest log as before with `retry=N/MAX` appended for context. Set to 0 to log every wait.

**LOW — extra blank lines at `:698` and `:790`**
- Both fixed.

**LOW — opaque `(yrru-mix3)` reference**
- Removed; docstring now cites PR #124.

**HIGH (tests) — flood before any yield**
- Added `test_iter_with_flood_retry_handles_flood_before_any_yield`. Removed the dead `fake_iter` from the original resume test.

**HIGH (tests) — multiple consecutive floods**
- Added `test_iter_with_flood_retry_survives_consecutive_floods` (3 in a row before success) and `test_iter_with_flood_retry_resets_counter_on_progress` (yield-then-flood pattern past the cap, proving the reset).

**MEDIUM (tests) — `max_id` preservation across retries**
- Added `test_iter_with_flood_retry_preserves_max_id_kwarg`.

**MEDIUM (tests) — overly broad `_fake_db` autouse fixture**
- Narrowed from `autouse=True` to an opt-in `fake_db` fixture. `test_config_kwargs_include_flood_sleep_threshold_zero` no longer pays the import-reload cost.

**Bonus tests**
- `test_iter_with_flood_retry_gives_up_after_max_retries` covers the new cap behavior end-to-end.
- `test_iter_with_flood_retry_caps_sleep_seconds` covers `MAX_FLOOD_WAIT_SECONDS` clamping.
- `test_iter_with_flood_retry_suppresses_short_wait_logs` covers `FLOOD_WAIT_LOG_THRESHOLD`.

Ready for another look.

GeiserX:
Thanks for the contribution, @tondeaf! Merged your PR and shipped a follow-up in #130 (v7.6.2) that hardens the retry logic further:

- `call_with_flood_retry()` for non-iterator API calls (`get_me`, `get_dialogs`)
- Negative `e.seconds` clamping
- `reverse=True` validation to prevent silent data corruption
- Defensive `FLOOD_WAIT_LOG_THRESHOLD` env parsing
- 6 additional edge-case tests (16 total)

Released as [v7.6.2](https://github.com/GeiserX/Telegram-Archive/releases/tag/v7.6.2).

Issue #149, opened 2026-05-07
[Feature]: Establish hierarchical structure under media/_shared


yarikoptic:
### Feature Category

Backup Process

### Problem Statement

In my initial attempt to backup my telegram I gathered

```
$> /bin/ls -1 ses-onerussian/backups/media/_shared | wc -l
436686
```
files flat (I think) in that folder!  That has various performance disadvantages for filesystems and VCS (git) if used.

### Proposed Solution

  I would advise to follow strategy employed by various tools like git, to introduce 1 or two levels of folders , like we have in git one level

```
(git)smaug:/mnt/btrfs/onerussian.com/sourcedata/src-telegram[master]git
$> ls -l .git/objects | head -n 3
total 0
drwxrwxr-x 1 yoh yoh  76 May  3 02:36 00/
drwxrwxr-x 1 yoh yoh 152 May  3 02:36 02/

$> ls -l .git/objects/00 | head -n 3
total 12604
-r--r--r-- 1 yoh yoh 12904252 May  2 17:43 5095d1a17b8d122461bf0e93692d32874aca80

$> ls -l .git/objects | nl | tail   
   178  drwxrwxr-x 1 yoh yoh 304 May  3 02:39 f6/
   179  drwxrwxr-x 1 yoh yoh  76 May  6 11:13 f7/
   180  drwxrwxr-x 1 yoh yoh  76 May  3 02:39 f8/
   181  drwxrwxr-x 1 yoh yoh  76 May  3 02:39 f9/
   182  drwxrwxr-x 1 yoh yoh  76 May  3 02:39 fb/
   183  drwxrwxr-x 1 yoh yoh 152 May  3 02:39 fc/
   184  drwxrwxr-x 1 yoh yoh 152 May  3 02:39 fd/
   185  drwxrwxr-x 1 yoh yoh  76 May  3 02:39 fe/
   186  drwxrwxr-x 1 yoh yoh  34 May  3 02:40 info/
   187  drwxrwxr-x 1 yoh yoh 296 May  3 02:35 pack/
```

and in git-annex 2

```
$> ls -l .git/annex/objects | nl | tail -n 2
  1024  drwxrwxr-x 1 yoh yoh 1396 May  2 21:39 zx/
  1025  drwxrwxr-x 1 yoh yoh 1380 May  2 21:38 zz/

$> ls -l .git/annex/objects/zz | nl | tail -n 2
[...]

PR #156, opened 2026-05-15
fix: include symlinks in sharding migration

GeiserX:
## Summary
- Removes the explicit symlink exclusion from `migrate_shared_media()`, addressing @yarikoptic's feedback in #149
- Symlinks with reachable targets (git-annex on host) now get sharded like regular files — the user's next `git commit` auto-fixes relative symlink targets via git-annex's pre-commit hook
- Symlinks with unreachable targets (Docker without `.git/`) are naturally skipped because `_compute_hash()` returns `None` on `OSError`
- No behavior change for non-git-annex users (regular files were never affected by the exclusion)

## Test plan
- [x] New test: `test_migrates_symlinks_with_reachable_targets` — verifies symlinks get sharded when target is readable
- [x] New test: `test_skips_symlinks_with_unreachable_targets` — verifies Docker case (broken targets) still skips gracefully
- [x] All existing sharding tests continue to pass (21/21)

Ref: https://github.com/GeiserX/Telegram-Archive/issues/149#issuecomment-4455495427

## Summary by CodeRabbit

* **Bug Fixes**
  * Symlinked media files are now included in the shared media migration process and properly moved to the sharded directory structure alongside regular files. Broken symlinks are handled appropriately.

* **Tests**
  * Added test coverage for symlink migration scenarios, including validation of reachable target conversion and handling of unreachable (broken) symlinks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GeiserX/Telegram-Archive/pull/156)

### DIFF
--- a/src/migrate_shared_media.py
+++ b/src/migrate_shared_media.py
@@ -41,16 +41,14 @@ def migrate_shared_media(media_path: str) -> int:
     if os.path.exists(marker):
         return 0
 
-    # Only process regular files directly in _shared/.
-    # Symlinks are excluded: moving a symlink to a subdirectory breaks its
-    # relative target (e.g. git-annex pointers like ../../.git/annex/objects/...).
-    # They remain in flat _shared/ and are found via the fallback in
-    # resolve_shared_file_path().
-    # .part files are excluded: they are in-progress downloads.
     flat_files = []
     try:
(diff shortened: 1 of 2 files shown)


yarikoptic:
Thank you!

Issue #175, opened 2026-05-29
[Bug]: Weird File with random numerical extensions and .part


zey0n:
### Version

Unknown

### Deployment Method

Docker Compose

### Database Type

PostgreSQL

### Bug Description

I was inspecting the saved files from an old chat, and found many files with name like fillename.mp4.<random-numbers> and fillename.mp4.part

--

Not sure when this error occurred or when these files are from, but I could see them. So maybe it's in the latest version or maybe it happened in a previous version, but just report.ing it here. 

--

I tried renaming some of those files and removed the random numbers from the extension and they actually worked. So that means those files were properly downloaded, but file extension had some sort of an issue.

### Expected Behavior

Files should have proper extension.

### Steps to Reproduce

-

### Logs

```shell

```

### Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [ ] I am using the latest version

GeiserX:
Thanks for reporting, will take a look around that. But I would love whatever more context you can give me.
Also make sure you're running also latest version of the software

zey0n:
I don't have more information on this. but let me monitor some new files getting downloaded from today

zey0n:
I was able to find the issue. This is latest version and issue with random with some media only in a chat. Attaching recent screenshots. Weird file extensions.

<img width="453" height="573" alt="Image" src="https://github.com/user-attachments/assets/231de5cf-f303-45f6-b899-640fc275eb79" />

<img width="644" height="472" alt="Image" src="https://github.com/user-attachments/assets/7bba9771-389e-4a04-b5be-dc1f117e0f87" />

GeiserX:
Fixed in **v7.11.3** (PR #176). Thank you @zey0n for the thorough report — and especially for confirming that renaming the files back made them open correctly. That detail was the key clue: it proved the file *contents* were intact and only the on-disk name was wrong, which pointed straight at the download-finalize step.

## What was happening

Downloads are written to a unique temporary path so concurrent downloads never collide: `{filename}.{pid}.{taskid}.part`. Telethon's `_get_proper_filename` treats the trailing `.part` as the file extension and hands the temp path back verbatim. The old finalize step stripped only the `.part` suffix, leaving names like `video.mp4.7.140234567890` on disk.

## The fix

1. **Forward fix** — finalize now renames the finished download to its intended clean filename regardless of what Telethon returns, so no new files can be corrupted.
[...]

PR #180, opened 2026-06-01
feat: decoupled async media download queue with concurrent downloads

smbdspk:
## Summary

This is a change I mostly implemented for myself to export media-heavy channels.
Rewrites the backup pipeline from sequential single-threaded processing to a producer-consumer architecture with concurrent media downloads. Synchronously downloading everything in a single thread is too slow and hits the speed limit on a single download thread for Telegram.
I also expanded on the delays by introducing a configurable min/max delays plus jitter, since Telegram-incurred throttle can become overly aggressive with concurrent downloads.
Since I ran a few passes on my DB before the async implementation was fully finished, I had to add the gap-filling functionality in case message batches got committed with failed separate messages.

### Key changes

- **Async download queue**: Messages are fetched by a producer and dispatched to N worker coroutines (`CONCURRENCY_LIMIT`, default 4) for concurrent media download, decoupling message iteration from media I/O
- **Exponential backoff with jitter**: Configurable `BACKOFF_MIN/MAX_SECONDS` for FloodWaitError and transient network errors (`TimeoutError`, `ConnectionError`, `OSError`, `RPCError`), replacing the flat +1s retry
- **Watermark checkpointing**: `_safe_checkpoint_id()` tracks in-flight message IDs to prevent sync cursor over-advancement during concurrent processing — crash-safe by design
- **Memory-bounded pipeline**: Queue `maxsize=1000`, backpressure when in-flight IDs exceed threshold, per-file download timeouts via `DOWNLOAD_TIMEOUT_SECONDS`
- **FileReferenceExpiredError handling**: Auto-refreshes expired media references with a 3-retry loop
- **Gap-fill mode** (opt-in via `FILL_GAPS=true`): Detects and recovers skipped messages from interrupted or partially-failed backups using SQL `LAG()` window functions. Includes trailing-gap recovery for over-advanced sync cursors

### New configuration variables

| Variable | Default | Description |
|---|---|---|
| `CONCURRENCY_LIMIT` | `4` | Max concurrent download workers per chat |
| `PRESERVE_ORDER` | `true` | Commit messages in Telegram ID order |
| `BACKOFF_MIN_SECONDS` | `2.0` | Minimum exponential backoff delay |
| `BACKOFF_MAX_SECONDS` | `300.0` | Maximum exponential backoff cap |
| `FILL_GAPS` | `false` | Enable gap detection and recovery |
| `GAP_THRESHOLD` | `50` | Minimum gap size to trigger fill |
| `DOWNLOAD_TIMEOUT_SECONDS` | `3600` | Per-file download timeout |
| `FLOOD_WAIT_LOG_THRESHOLD` | `10` | Suppress FloodWait logs below this (seconds) |
| `MAX_FLOOD_RETRIES` | `5` | Maximum retries for FloodWait limits |
| `MAX_FLOOD_WAIT_SECONDS` | `3600` | Abort if FloodWait exceeds this |

All defaults preserve prior sequential behavior. Setting `CONCURRENCY_LIMIT=1` + `PRESERVE_ORDER=true` replicates the old pipeline exactly, just through the new async machinery.

## Type of Change

- [ ] Documentation update

> **Note:** While the internal architecture of `_backup_dialog()` is completely rewritten from a sequential loop to a concurrent producer-consumer pattern, all defaults preserve existing behavior. This is a significant implementation change but not a breaking API/behavior change.

## Database Changes

- [x] No database changes

New DB adapter methods (`detect_message_gaps`, `detect_trailing_gaps`, `reset_sync_cursor`, `get_chats_with_messages`) operate on pre-existing tables (`messages`, `sync_status`, `chats`).

## Data Consistency Checklist

Database code structure was not modified - only new read queries and a cursor reset were added.

## Testing

### Coverage notes

Good test coverage for flood retry, exponential backoff, and gap-fill SQL logic. The concurrent producer-consumer core (`_producer_loop`, `_worker_loop`, `_safe_checkpoint_id`) is tested manually but lacks automated unit tests.

## Security Checklist

## Deployment Notes

No special deployment steps. Docker image rebuild required for the new code paths. All new features are controlled by environment variables with backward-compatible defaults.

## Summary by CodeRabbit

* **New Features**
  * Configurable bounded exponential backoff with jitter for flood-wait and transient retries; new env knobs for backoff bounds
  * New env knobs: media download timeout, gap-filling controls, DB tuning, and master notifications toggle

* **Bug Fixes**
  * More robust media download with timeouts, refresh/retry for expired references, and hardened concurrent symlink handling
  * Retries for transient DB lock errors and stricter cached-stats parsing

* **Documentation**
  * Expanded example environment file with tuning guidance

* **Tests**
  * Updated and expanded tests to cover backoff, jitter, path robustness, and platform-specific symlink behavior

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,9 +44,19 @@ DOWNLOAD_MEDIA=true
 # Files larger than this will be skipped
 MAX_MEDIA_SIZE_MB=100
 
+# Timeout for media downloads (in seconds). 0 disables the timeout.
+# Default is 3600 (60 minutes).
+# DOWNLOAD_TIMEOUT_SECONDS=3600
+
 # Messages processed per database batch
 # Higher values = faster but more memory usage
 BATCH_SIZE=100
+
+# Detect and fetch missing messages in sequence gaps
+# FILL_GAPS=false
+
+# Number of messages missing in a sequence to consider it a gap
+# GAP_THRESHOLD=50
 
 # How often to save backup progress (every N batch inserts)
 # 1 = checkpoint every batch (safest, best crash recovery)
@@ -85,6 +95,16 @@ LOG_LEVEL=INFO
 # Waits shorter than this are routine and logged at DEBUG instead of WARNING
 # Set to 0 to log every flood-wait
 # FLOOD_WAIT_LOG_THRESHOLD=10
+
+# Maximum retries for FloodWait limits
+# MAX_FLOOD_RETRIES=5
+
+# Maximum seconds to wait during a FloodWait before aborting
+# MAX_FLOOD_WAIT_SECONDS=3600
+
+# Exponential backoff for transient connection errors
+# BACKOFF_MIN_SECONDS=2.0
+# BACKOFF_MAX_SECONDS=300.0
 
 # Session name (use different names for multiple accounts)
 SESSION_NAME=telegram_backup
@@ -159,6 +179,13 @@ VERIFY_MEDIA=false
 # Option 2: Individual settings
 DB_TYPE=sqlite
 DB_PATH=/data/backups/telegram_backup.db
+
+# Timeout for SQLite operations (seconds)
+# Increase if you see "database is locked" errors on slow disks
+# DATABASE_TIMEOUT=60.0
+
+# Print all SQL queries to console (for debugging)
+# DB_ECHO=false
 
 # PostgreSQL settings (when DB_TYPE=postgresql)
 # POSTGRES_HOST=localhost
@@ -240,6 +267,9 @@ DB_PATH=/data/backups/telegram_backup.db
 # ==============================================================================
 # NOTIFICATIONS
 # ==============================================================================
+
+# Master switch to enable push notifications
+# ENABLE_NOTIFICATIONS=false
 
 # Push notification mode: off, basic (in-browser only), full (Web Push)
 # PUSH_NOTIFICATIONS=basic
--- a/src/config.py
+++ b/src/config.py
@@ -113,6 +113,8 @@ class Config:
         self.backup_path = os.path.abspath(os.getenv("BACKUP_PATH", "/data/backups"))
         self.download_media = os.getenv("DOWNLOAD_MEDIA", "true").lower() == "true"
         self.max_media_size_mb = int(os.getenv("MAX_MEDIA_SIZE_MB", "100"))
+        # Timeout for media downloads (seconds). 0 disables the timeout.
+        self.download_timeout_seconds = int(os.getenv("DOWNLOAD_TIMEOUT_SECONDS", "3600"))
 
         # Batch processing configuration
         self.batch_size = int(os.getenv("BATCH_SIZE", "100"))
@@ -125,6 +127,10 @@ class Config:
         # Increase this if you experience "database is locked" errors (e.g., on Unraid/slow disks).
         # Default increased to 60s for better resilience with concurrent access (backup + web viewer).
         self.database_timeout = float(os.getenv("DATABASE_TIMEOUT", "60.0"))
+
+        # Backoff intervals for connection/timeout retries.
+        self.backoff_min = float(os.getenv("BACKOFF_MIN_SECONDS", "2.0"))
+        self.backoff_max = float(os.getenv("BACKOFF_MAX_SECONDS", "300.0"))
 
         # =====================================================================
         # CHAT FILTERING - Two Modes
--- a/src/connection.py
+++ b/src/connection.py
@@ -14,6 +14,7 @@ Architecture:
 import asyncio
 import logging
 import os
+import random
 import shutil
 import sqlite3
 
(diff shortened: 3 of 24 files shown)


GeiserX:
Hi @smbdspk — first off, thank you for this. It's a serious, well-documented piece of work, and the fact that you ran it against your own media-heavy channels and then thought carefully about checkpointing, backpressure, timeouts, and gap recovery shows real care. A lot of what's in here is genuinely valuable, and I want to make sure none of it gets lost. I spent a good while reviewing it (and cross-checking the Telethon internals), so let me share where I landed — and a path I think gets your improvements merged quickly.

## The part I'd like us to reconsider: concurrency on one shared client

The architecture is built on the premise that "synchronously downloading in a single thread hits Telegram's per-thread speed limit, so N workers go faster." I dug into this and I don't think the premise holds for the specific shape here — **N worker coroutines all calling `download_media` on a single shared `TelegramClient`**:

- A single Telethon client uses **one MTProto connection per DC**, and `download_media` is **not** parallel. From the official FAQ: *"the library does not download or upload files in parallel… the limiting factor in the long run are `FloodWaitError`, and using parallel download or uploads only makes them occur sooner."* (https://docs.telethon.dev/en/stable/quick-references/faq.html)
- Telegram's rate limit is **per-account/per-DC, not per-connection**. So concurrent requests on one account don't raise the ceiling — they spend the flood budget faster. Telethon's author makes this explicit (https://github.com/LonamiWebs/Telethon/issues/4376, and a dedicated "concurrent download with multiple workers" request was closed as not-planned: https://github.com/LonamiWebs/Telethon/issues/4550).
- This actually matches your own observation in the PR description — that throttle "can become overly aggressive with concurrent downloads," which is what prompted the jitter. That's the tell: the concurrency creates the flood pressure that the backoff then has to absorb.

The community-standard way to genuinely speed up Telethon downloads is different: **chunk a single file across multiple connections** (the FastTelethon `ParallelTransferrer` pattern — https://gist.github.com/painor/7e74de80ae0c819d3e9abcf9989a8dd6, capped at ~20 senders), plus making sure **`cryptg` is installed** (often the single biggest win — pure-Python AES is the real bottleneck on many setups). Tools known for fast Telegram downloads (e.g. `tdl`) parallelize *within* a file and pool connections per DC, rather than running many files over one client.

## A couple of correctness concerns I'd want to resolve before concurrency lands

These are the ones that worry me for a backup tool, where a silent skip is unrecoverable:

1. **Sync-cursor over-advancement.** With `PRESERVE_ORDER=false`, a message whose download ultimately fails is dropped from in-flight tracking, but `running_max_id` still includes it — and the final checkpoint persists `running_max_id`. That can advance the cursor past a message that never got saved. `_recover_trailing_gaps` is clearly meant to catch this, but it keys on `MAX(messages.id)`, so it can't see the "message row committed, media missing, cursor moved past it" case.
2. **`CONCURRENCY_LIMIT=1` isn't quite the old path.** Even at 1, flow now goes through the queue + the polling committer loop, and the **default is 4** — so existing deployments would change behavior on upgrade rather than staying on the proven sequential path.
3. **The concurrent core is untested.** The producer/worker/committer/checkpoint logic (the riskiest part) is currently exercised only by mocks with synchronous side effects, so race/crash interleavings aren't covered.

There's also relevant history: the sequential design was a deliberate, incident-driven choice (#77, after a large backup repeatedly restarted from zero), and #175 was a corruption cascade rooted in concurrency-oriented temp paths that took four follow-up PRs to fully heal. So I'm being extra cautious specifically around the sync/commit path.

## What I'd love to merge right away

Honestly, a big chunk of this PR is excellent and **independent of the concurrency question**. If you're open to it, I'd be glad to take these in a focused PR (and credit you in the changelog):

- The exponential backoff + jitter for `FloodWaitError` and transient errors
- The `FileReferenceExpiredError` auto-refresh loop
- `DOWNLOAD_TIMEOUT_SECONDS` per-file timeouts
- The additional `@retry_on_locked` coverage and the symlink `FileExistsError` hardening
- The Python-3.14 `except (A, B)` fixes

## Suggested path

1. Split out the resilience improvements above → I'll review and merge those quickly.
2. For speed, if you're up for it, swap the file-level workers for FastTelethon-style **per-file chunked download** (single file, multiple senders, ≤20) + confirm `cryptg` is present. That's the approach that actually moves the bandwidth ceiling without multiplying floods.
3. If you can share a quick benchmark (wall-clock on a media-heavy chat at `CONCURRENCY_LIMIT=1` vs `4`), that would settle the throughput question with real numbers — I'd genuinely like to see them.

None of this is a knock on the effort — it's a lot of thoughtful engineering, and I want to land the parts that help everyone while keeping the backup path bulletproof. Let me know which direction you'd prefer and I'm happy to help with the split.

FisHlaBsoMAN:
Lol

smbdspk:
OK, done. If everything looks good expect a follow-up PR with parallel chunked downloads; or, alternatively, I'd rather open an issue and let your agents handle this, since you most definitely have considerably more context on the project.

GeiserX:
Merged — thank you for the rework. Dropping the concurrency rewrite and keeping the resilience improvements turned this into a focused, low-risk change: the backup path stays sequential and crash-safe, no orphaned config left behind, and the full suite is green (1794 tests, ruff + format clean).

I'm folding four small follow-up tidy-ups into `main` myself in a quick follow-up PR, so there's nothing further needed from you:

1. Removed the unused `config.backoff_min` / `config.backoff_max` (the backoff math uses the module-level globals).
2. Restored `except (json.JSONDecodeError, TypeError)` in `adapter.py` so non-str cached stats stay caught.
3. Moved the `.part` cleanup inside the download retry loop so partials don't linger between attempts.
4. Used the resolved `timeout` local in the timeout log line instead of re-reading config.

You'll be credited in the v7.12.0 changelog. And yes — please open an **issue** (not a PR) for the parallel/chunked downloads so we can align on the FastTelethon per-file chunking + `cryptg` approach before any code. Thanks again for the contribution.

PR #191, opened 2026-06-10
fix(backup): preserve service action metadata in raw_data

### LINKED ISSUE #190: [Bug]: Historical backfill drops service-action metadata that the live listener preserves
### What happens

`_process_message` (src/telegram_backup.py) stores service messages during historical backfills but discards `message.action` entirely: `raw_data` is enriched for albums, forwards, post authors and polls, while service actions are dropped.

The live listener has preserved this metadata since v6.0.0 (`raw_data.service_type` / `action_type` / `new_title`, src/listener.py). The two ingestion paths are inconsistent: an event captured live keeps its payload; the same event reached through a backfill loses it, and the loss is permanent.

### Why it matters

Forum topic history is the visible casualty. `forum_topics.title` is overwritten on rename (upsert in src/db/adapter.py), and the `MessageActionTopicCreate` / `MessageActionTopicEdit` service messages are stored with no action payload. After a backfill there is no way to audit when a topic was created or renamed, or what it was called before.

### Reproduction (synthetic forum)

1. In a test forum supergroup: create a topic, post in it, rename it.
2. Run a historical backup of that chat.
3. Inspect the DB: the service rows exist (empty text) but raw_data carries no action info; the previous topic title is gone.

### Proposed fix (minimal, no schema change)

Mirror the listener convention in `_process_message`:

- `raw_data.service_type = "service"`;
- `raw_data.action_type` = normalized action class name (`topic_create`, `topic_edit`, `chat_edit_title`, ...);
- `raw_data.new_title` via the existing `_text_with_entities_to_string` helper, only when the action carries a title.

Topic history becomes reconstructible with existing columns: `topic_create` links via `messages.id == forum_topics.id`; `topic_edit` via `reply_to_top_id`. No migration (raw_data is already JSON), no new logs, no viewer changes.

### Validated

With the fix (+4 unit tests following the existing raw_data test pattern; full suite 1847 passed on Python 3.14; ruff clean), a fresh backfill of a synthetic test forum recovered all 4 service actions with their titles, fully linkable — including a topic rename performed before any backup existed.

I have the PR ready and I am opening it right after this issue.

cdelalama:
## Summary

- Fixes #190: historical backfills (`_process_message`) now preserve `message.action` metadata in `raw_data`, restoring parity with the live listener's convention (`service_type` / `action_type` / `new_title`, in place since v6.0.0).
- Forum topic history becomes auditable and reconstructible with existing columns: `topic_create` links via `messages.id == forum_topics.id`; `topic_edit` via `reply_to_top_id`. `new_title` goes through the existing `_text_with_entities_to_string` helper.

## Type of Change

- [ ] Documentation update

## Database Changes

- [x] No database changes

## Data Consistency Checklist

- [x] All `chat_id` values use marked format (via `_get_marked_id()`) — no chat_id handling changed
- [x] All datetime values pass through `_strip_tz()` before DB operations — no datetime handling changed
- [x] INSERT and UPDATE operations handle the same fields identically — no DB operations changed

## Testing

- [x] Tests pass locally (`python -m pytest tests/ -v`) — 1847 passed on Python 3.14
- [x] Manually tested in development environment — fresh backfill of a synthetic test forum recovered all 4 service actions (2x topic_create, 1x topic_edit, 1x channel_migrate_from) with titles, fully linkable, including a topic rename performed before any backup existed

## Security Checklist

- [x] User input properly validated/sanitized — values come from Telethon objects, titles through the existing TextWithEntities helper
- [x] Authentication/authorization properly checked — N/A, no auth paths touched

## Deployment Notes

- No image or config changes needed; existing archives are unaffected. Already-backfilled service rows gain metadata only if their history is re-fetched.
- Naming is easy to adjust if you prefer different keys (e.g. `action_title` instead of `new_title`, or the listener's curated vocabulary for `action_type`).

## Summary by CodeRabbit

* **New Features**
  * Backups now capture and preserve service event metadata for forum topic operations (creations and renames), ensuring these events are tracked with full context in historical records.

### DIFF
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -8,6 +8,7 @@ import base64
 import logging
 import os
 import random
+import re
 from datetime import UTC, datetime
 
 from telethon import TelegramClient
@@ -287,6 +288,17 @@ async def iter_messages_with_flood_retry(client, entity, *, min_id=0, **kwargs):
                     MAX_FLOOD_RETRIES,
                 )
             await asyncio.sleep(sleep_duration)
+
+
+def _service_action_type(action: object) -> str:
+    """Normalize a Telethon MessageAction class name to snake_case.
+
+    Examples: MessageActionTopicCreate -> "topic_create",
+    MessageActionTopicEdit -> "topic_edit",
+    MessageActionChatEditTitle -> "chat_edit_title".
+    """
+    name = type(action).__name__.removeprefix("MessageAction")
+    return re.sub(r"(?<!^)(?=[A-Z])", "_", name).lower()
 
 
 class TelegramBackup:
@@ -1411,6 +1423,19 @@ class TelegramBackup:
             "is_outgoing": 1 if message.out else 0,
             "is_pinned": 1 if getattr(message, "pinned", False) else 0,
         }
+
+        # Preserve service-action metadata (e.g. forum topic creations and
+        # renames) so historical backfills keep parity with the listener's
+        # raw_data convention (service_type / action_type, since v6.0.0).
+        # Without this, service events are stored without their payload and
+        # the information is irrecoverable once the history is archived.
+        action = getattr(message, "action", None)
+        if action is not None:
+            message_data["raw_data"]["service_type"] = "service"
+            message_data["raw_data"]["action_type"] = _service_action_type(action)
+            action_title = getattr(action, "title", None)
+            if action_title is not None:
+                message_data["raw_data"]["new_title"] = self._text_with_entities_to_string(action_title)
 
         # Capture grouped_id for album detection (multiple photos/videos sent together)
         if message.grouped_id:
--- a/tests/test_telegram_backup.py
+++ b/tests/test_telegram_backup.py
@@ -12,6 +12,8 @@ from unittest.mock import AsyncMock, MagicMock
 from telethon.tl.types import (
     Channel,
     Chat,
+    MessageActionTopicCreate,
+    MessageActionTopicEdit,
     MessageMediaContact,
     MessageMediaDocument,
     MessageMediaGeo,
@@ -366,6 +368,9 @@ class TestBackupCheckpointing(unittest.TestCase):
         # topic-skip guard in _backup_dialog doesn't accidentally filter
         # every message via MagicMock truthiness.
         msg.reply_to = reply_to
+        # Explicitly None so the service-action capture in _process_message
+        # is not triggered by MagicMock truthiness.
+        msg.action = None
         return msg
 
     def test_checkpoint_after_every_batch(self):
@@ -1164,6 +1169,9 @@ class TestProcessMessage(unittest.TestCase):
         msg.media = None
         msg.reactions = None
         msg.post_author = None
+        # Explicitly None so the service-action capture in _process_message
+        # is not triggered by MagicMock truthiness.
+        msg.action = None
         return msg
 
     def test_basic_text_message(self):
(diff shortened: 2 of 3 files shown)


GeiserX:
Thanks for this, @cdelalama — it's a genuinely well-crafted PR. 🙏 You opened a precise issue first, scoped the fix to the actual seam (`_process_message`), avoided a schema migration, anticipated the `MagicMock` truthiness pitfall by pinning `msg.action = None` in the touched factories, added four tests covering create / edit-with-title / edit-without-title / regular-message, and the commit message even pre-states the linkage invariants. CodeRabbit found nothing actionable; neither did we on correctness.

I ran a deep multi-agent review anyway (Telethon-API, logic, tests, architecture, simplicity) and verified the key claims against the **installed Telethon 1.43.2** source and an empirical repro rather than from memory. Sharing the findings for transparency — **none block merge**; the verdict is **approve, with a couple of small follow-ups you might consider.**

## Verified correct ✅
- **The `getattr(message, "action", None) is not None` guard fires only for service messages.** Telethon's own docstring (`tl/custom/message.py`, identical in 1.41.0 and 1.43.2): *"the message action object … for `MessageService` instances, which will be None for other types."* On a regular `Message`, `.action` is a real attribute defaulting to `None` — confirmed by construction, no truthiness trap in production.
- **`new_title` only-when-`.title is not None` is correct.** `MessageActionTopicEdit` types `title` as `flags.0?string` (optional); a close/hide/icon-only edit has `.title is None`. Verified against the TL schema (layer 224) and by constructing real objects — your `test_topic_edit_without_title_has_no_new_title` nails this.
- **Class names are stable across Telethon 1.41–1.43** (no renames/removals), so the name-derived `action_type` is robust; new action classes only add, never break.
- **Idempotent & self-healing.** Backfill funnels through `insert_messages_batch` → `on_conflict_do_update(... set_=values)` including `raw_data`, so a re-fetch *overwrites* old service rows with the new metadata. Your deployment note is accurate (and slightly understated — re-fetched rows are healed, not just new ones).
- **No exception risk, no key collisions.** Both helpers are total over real inputs; the three keys are disjoint from the later `grouped_id`/`forward`/`poll` blocks; chat_id/datetime normalization is untouched (you only write into the in-memory `raw_data`). Full suite **1847 passed**, ruff clean.

## Worth a look (non-blocking)

**1. (nit, test hygiene) One mock factory still omits `msg.action`.** You correctly pinned `msg.action = None` in both `_make_message` factories in `tests/test_telegram_backup.py`, but `tests/test_telegram_backup_extended.py:47` was missed. Six tests there call the *real* `_process_message` (lines 1631, 1651, 1661, 1688, 2219, 2241), so their bare `MagicMock` now silently injects junk into `raw_data`. Reproduced:
```
service_type = 'service'
action_type  = 'magic_mock'
new_title    = "<MagicMock name='mock.action.title' id=...>"
```
They still pass (their assertions only check `reactions`/`poll`, never these keys), and **production is unaffected** (a real `.action` is `None`). It's purely a test-fixture smell. One-line fix: add `msg.action = None` near `msg.post_author = None` at `test_telegram_backup_extended.py:64`. A direct unit test for `_service_action_type` would also be nice — it's currently only exercised transitively.

**2. (minor) "Parity with the listener" is structural, not value-level.** The `raw_data` *keys* now match the listener — but the *values* are a different vocabulary. The listener derives `action_type` from `events.ChatAction` (`title_changed`, `user_added`, …); your backfill derives it from the `MessageAction` class name (`chat_edit_title`, `chat_add_user`, …). So the *same* group-title change is `title_changed` live vs `chat_edit_title` on backfill. **Inert today** — exhaustive grep confirms nothing in the repo reads `action_type`/`new_title` (the viewer gates only on `service_type === 'service'` and renders `{{ msg.text }}`). It's a latent trap for a future consumer (e.g. the external MCP server, or a query like "all title changes"). Suggestion: just reword the comment to "structural parity (keys)" and/or add a note flagging the two vocabularies. No code change needed.

**3. (minor, cosmetic) Acronym class names snake-case oddly.** `re.sub(r"(?<!^)(?=[A-Z])", "_", name)` splits consecutive capitals letter-by-letter: `MessageActionSetMessagesTTL` → `set_messages_t_t_l`. None of the topic/title actions you target are affected (those are all clean), and nothing reads the value, so it's deterministic-but-ugly only. Worth a one-line docstring note so it doesn't surprise a future reuse.

**4. (FYI, out of scope) Backfilled service messages render as empty bubbles.** `_process_message` sets `text = message.text or ""`, and a `MessageService` carries its payload in `.action`, not `.message` — so the viewer's `{{ msg.text }}` shows an empty pill. This is **pre-existing** (not introduced by this PR — you only *add* `raw_data`) and explicitly outside #190's "auditability via raw_data" scope, but since this PR now tags those rows `service_type='service'`, it's the natural next step: a follow-up could synthesize display text (mirroring the listener's `service_text`) or have the viewer fall back to `action_type`/`new_title` when `text` is empty.

## One small thing you *could* fold in now
`_service_action_type` is a pure function and a natural fit for `src/message_utils.py`, right next to `extract_topic_id` (the existing shared helper both the listener and backfill already import). Keeping `new_title` extraction in `telegram_backup.py` is fine since it needs the instance method. Totally optional.

Net: clean, correct, safe to merge as-is — the only thing I'd genuinely suggest before merge is the one-line test-fixture pin (#1). Everything else is polish or a follow-up. Really nice work, and thanks again for contributing this. 🙌

GeiserX (approved):

PR #199, opened 2026-06-25
feat(messages): add soft deletion mode

charys117:
## Summary

Add configurable soft deletion support for archived Telegram messages.

This introduces `DELETION_MODE=soft|hard` for both real-time deletion listener handling and batch deletion/edit sync:

- `hard` keeps the existing behavior and removes archived messages.
- `soft` keeps the original archived message and marks it as deleted.
- The viewer displays deleted messages with a `deleted` metadata label.
- Messages that are both edited and deleted display `edited deleted`.

## Motivation

Previously, enabling deletion handling could mirror Telegram deletions by removing messages from the local archive. That made it hard to preserve archive history while still recording that a message was deleted upstream.

This change allows users to keep the original message content while recording deletion state.

## Changes

- Add `DELETION_MODE` config with default `hard` for backwards compatibility.
- Add `messages.is_deleted` and `messages.deleted_at` columns via Alembic migration `014`.
- Add `mark_message_deleted()` DB adapter method.
- Preserve soft-delete markers during message upserts so retries, gap fill, or reprocessing do not clear deletion state.
- Apply `DELETION_MODE` in the real-time deletion listener and `SYNC_DELETIONS_EDITS=true` batch sync.
- Update WebSocket delete handling so soft deletes update the visible message instead of removing it.
- Update viewer metadata rendering: edited only = `edited`, deleted only = `deleted`, both = `edited deleted`.
- Document `DELETION_MODE` in README and `.env.example`.
- Pass `DELETION_MODE` through `docker-compose.yml`.

## Compatibility

The default remains `DELETION_MODE=hard`, so existing deployments that already enabled deletion handling keep the previous archive behavior.

Users who want to preserve deleted messages should explicitly set:

```env
DELETION_MODE=soft
```

For local direct runs against an existing SQLite DB, run migrations before starting the app:

```bash
uv run alembic upgrade head
```

Docker entrypoint-managed deployments already run Alembic migrations on startup.

## Testing

Ran:

```bash
uv run python -m py_compile src/config.py src/db/adapter.py src/db/models.py src/listener.py src/telegram_backup.py src/web/main.py alembic/versions/20260625_014_add_message_soft_delete.py

uv run --with ruff ruff check src/config.py src/db/adapter.py src/db/models.py src/listener.py src/telegram_backup.py src/web/main.py tests/test_config.py tests/test_listener.py tests/test_listener_extended.py tests/test_telegram_backup_extended.py tests/test_db_adapter_soft_delete_upsert.py alembic/versions/20260625_014_add_message_soft_delete.py

uv run --with pytest --with pytest-asyncio python -m pytest tests/test_config.py::TestListenerLogging tests/test_listener.py::TestEventHandlers tests/test_listener_extended.py::TestOnMessageDeletedRateLimiting tests/test_telegram_backup_extended.py::TestSyncDeletionsAndEdits tests/test_db_adapter_soft_delete_upsert.py -q

uv run --with pytest --with pytest-asyncio python -m pytest tests/test_web_main.py tests/test_web_routes.py -q

uv run --with pytest --with pytest-asyncio python -m pytest tests/test_db_adapter.py tests/test_db_adapter_soft_delete_upsert.py -q
```

Also smoke-tested SQLite Alembic upgrade to head and confirmed migration `014` creates `messages.is_deleted` and `messages.deleted_at`.

## Summary by CodeRabbit

* **New Features**
  * Added `DELETION_MODE` (`hard` or `soft`) to control how deleted Telegram messages are processed (default `hard`).
  * WebSocket delete events and the UI now support soft-deletion metadata, including a “deleted” indicator and message dimming.
* **Bug Fixes**
  * Deletion syncing and stored message updates now follow the selected deletion mode, preserving soft-deleted state on upserts.
* **Documentation**
  * Updated environment and README examples to explain `DELETION_MODE`, including the real-time listener behavior.
* **Tests**
  * Expanded test coverage for configuration parsing/validation, listener/backup deletion behavior, and soft-delete upsert & migration 014.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -173,6 +173,7 @@ ENABLE_LISTENER=false
 # Granular listener controls (only apply when ENABLE_LISTENER=true):
 # LISTEN_EDITS=true
 # LISTEN_DELETIONS=false
+# DELETION_MODE=hard  # hard=legacy delete from archive, soft=mark deleted and keep message
 # LISTEN_NEW_MESSAGES=true
 # LISTEN_NEW_MESSAGES_MEDIA=false
 # LISTEN_CHAT_ACTIONS=true
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ LISTEN_NEW_MESSAGES: "true"    # Save new messages instantly (default: true)
 
 **How it works:** stays connected to Telegram between scheduled backups, captures changes as they happen, and automatically reconnects if disconnected.
 
-**Backup protection:** `LISTEN_DELETIONS=false` is the safe default. Set `LISTEN_DELETIONS=true` only if you want to process deletion events. With the default `DELETION_MODE=hard`, deletions mirror Telegram and remove archived messages. Set `DELETION_MODE=soft` to keep the original archived message and show `deleted` in the message metadata.
+**Backup protection:** `LISTEN_DELETIONS=false` is the safe default. Set `LISTEN_DELETIONS=true` only if you want to process deletion events. With the default `DELETION_MODE=hard`, deletions mirror Telegram and remove archived messages. Set `DELETION_MODE=soft` to keep the original archived message and show `deleted` in the message metadata. Soft-deleted messages are retained in the archive — they remain counted in chat statistics and continue to appear in search and exports, flagged as `deleted`.
 
 **Alternative — batch sync:** set `SYNC_DELETIONS_EDITS=true` to check ALL backed-up messages on each scheduled run. This is expensive and slow, and uses the same `DELETION_MODE` behavior for deleted messages.
 
--- a/README.md
+++ b/README.md
@@ -285,7 +285,8 @@ The **Scope** column shows whether each variable applies to the backup scheduler
 | **Real-time Listener** | | | See [Real-time Listener](#real-time-listener) below |
 | `ENABLE_LISTENER` | `false` | B | **Master switch** — enables all `LISTEN_*` features below |
 | `LISTEN_EDITS` | `true` | B | Apply text edits in real-time |
-| `LISTEN_DELETIONS` | `false` | B | Mirror deletions from Telegram. Opt-in only; enabling this can delete archived messages |
+| `LISTEN_DELETIONS` | `false` | B | Process deletion events from Telegram. Opt-in only |
+| `DELETION_MODE` | `hard` | B | When deletions are processed: `hard` removes archived messages (legacy), `soft` keeps messages and marks them deleted |
 | `LISTEN_NEW_MESSAGES` | `true` | B | Save new messages in real-time between scheduled backups |
 | `LISTEN_NEW_MESSAGES_MEDIA` | `false` | B | Also download media immediately (vs. next scheduled backup) |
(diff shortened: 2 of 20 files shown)


charys117:
Addressed the CodeRabbit review items in the latest push:

- removed chat_id from the new soft-delete debug log
- included chat_id in edit/delete websocket payloads
- matched realtime edit/delete updates by chat_id + message_id on the frontend
- stamped pre-Alembic schemas with 014 when the soft-delete columns already exist
- added the missing return annotation on broadcast_message_delete

I intentionally left DELETION_MODE wired in docker-compose.yml. It controls batch sync behavior as well as listener deletion handling, so keeping it passed through lets .env configure soft vs hard deletion for both paths.

GeiserX:
Thanks a lot for this, @charys117 — a really well-scoped feature, and I appreciate that you shipped it with the Alembic migration, tests, and docs all included. 🙏

I ran a thorough multi-perspective review and the core design holds up well: the upsert soft-delete preservation (`_message_conflict_update_values`) is correct on both the SQLite and Postgres dialects, `func.coalesce` keeps the first deletion timestamp (nicely idempotent), migration 014 is metadata-only (no table rewrite), and the frontend `chat_id` matching is type-safe end-to-end.

I pushed a follow-up commit (`ee4db53`) on top of your branch with some hardening so it's merge-ready:

**Robustness**
- Added `@retry_on_locked` to `mark_message_deleted` (and `delete_message`). Without it, the real-time listener and the scheduled backup — which both write the same SQLite DB concurrently — can hit "database is locked" and silently drop a soft-delete (the listener swallowed that at debug while still counting it as applied).
- Wrapped the web entrypoint's `Config()` so an invalid `DELETION_MODE` surfaces a clear error instead of an opaque ASGI import-time crash-loop.

**Polish / project conventions**
- Dropped `chat_id` from the deletion debug log lines (we treat chat IDs as PII).
- The resolved-chat deletion path now counts and surfaces errors instead of swallowing them at debug, matching the other deletion path.
- Frontend: indexed the auto-refresh poll by id (avoids an O(latest × in-view) scan every 3s on long chats) and gave soft-deleted message bodies a muted style so they're visually distinct.
- Consolidated the duplicated `_utcnow_naive` helper into `message_utils`, a couple of small cleanups, and the `ruff format` reflows that were failing CI.

**Tests (+14)** for the upsert-preservation branches, coalesce-keeps-first, the `get_messages_sync_data` exclusion filter, migration 014 idempotency + downgrade, the listener notify payloads (soft + hard), and `DELETION_MODE` normalization.

Locally everything is green — `ruff check` + `ruff format --check` clean, and the full suite at 1877 passing. CI should follow.

I intentionally left the unused `broadcast_*` helpers and the entrypoint stamp ordering alone, since those are pre-existing and out of scope here.

Thanks again for the contribution! 🚀

GeiserX:
Merged! 🎉 Thanks again, @charys117 — soft-deletion mode is a great addition, and shipping it with the migration, tests, and docs made the review easy. This goes out in **v7.16.0** (cutting the release now). Appreciate the contribution! 🙏

PR #203, opened 2026-06-26
fix(media): refresh message on location unavailable

charys117:
## Summary

- Treat Telegram `LOCATION_NOT_AVAILABLE` media download errors as refreshable in the backup media path.
- Bypass generic stale-reference RPC retries for that error so the outer media retry loop can fetch a fresh message before retrying.
- Cover direct, deduplicated, stale-reference, non-refreshable RPC, missing-refresh, and final-retry failure paths with regression tests.

## Type of Change

- [ ] Documentation update

## Database Changes

- [x] No database changes

## Data Consistency Checklist

- [x] All `chat_id` values use marked format (via `_get_marked_id()`)
- [x] All datetime values pass through `_strip_tz()` before DB operations
- [x] INSERT and UPDATE operations handle the same fields identically

Not applicable: this PR does not change database reads or writes.

## Testing

- [ ] Manually tested in development environment

Commands run:

- `python3 -m py_compile src/telegram_backup.py tests/test_telegram_backup_extended.py`
- `/tmp/telegram-archive-test-venv/bin/python -m pytest tests/test_telegram_backup_extended.py -q`
- `/tmp/telegram-archive-test-venv/bin/python -m pytest tests/test_flood_wait_visibility.py -q`
- `/tmp/telegram-archive-test-venv/bin/python -m pytest tests/test_telegram_backup_extended.py tests/test_flood_wait_visibility.py --cov=src.telegram_backup --cov-report=term-missing -q`
- `/tmp/telegram-archive-test-venv/bin/python -m ruff check src/telegram_backup.py tests/test_telegram_backup_extended.py`
- `/tmp/telegram-archive-test-venv/bin/python -m ruff format --check src/telegram_backup.py tests/test_telegram_backup_extended.py`

## Security Checklist

This PR also avoids logging chat IDs or message IDs in the new media-location retry logs.

## Deployment Notes

- Docker image rebuild needed for deployments to pick up the backup retry behavior.
- No configuration, database migration, or manual repair step required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Telegram media reliability by detecting media “location not available/invalid” failures and refreshing the message before retrying the download.
  * Bounded media retry/backoff with jitter and per-operation timeouts to avoid excessive retry pacing; file-reference-expired cases retry immediately without added backoff.
  * More correct error propagation and cleanup: avoids corrupt partial `.part` files and stops refresh/retry for unrelated errors.

* **Tests**
  * Expanded coverage for media error classification, refresh/backoff timing, deduplicated vs non-deduplicated download flows, bounded timeout behavior, and partial-download cleanup.

### DIFF
--- a/src/media_errors.py
+++ b/src/media_errors.py
@@ -1,0 +1,42 @@
+"""Classification of refreshable Telegram media-download errors.
+
+Telegram's ``upload.GetFile`` can report a media file's storage location as
+temporarily unavailable or invalid. These are transient/per-file conditions:
+the documented response (mirrored by the official Telegram Desktop client) is to
+re-fetch the message for a fresh reference/location and retry, then leave the
+item for a later attempt rather than failing hard.
+
+Two shapes reach us through Telethon:
+
+* ``LOCATION_NOT_AVAILABLE`` has no generated Telethon exception class, so
+  ``rpc_message_to_error`` falls back by numeric code (400 -> ``BadRequestError``)
+  and preserves the raw server string in ``exc.message`` — match it by text.
+* ``LOCATION_INVALID`` *does* have a generated class (``LocationInvalidError``)
+  whose ``.message`` is the generic ``"BAD_REQUEST"`` — match it by type.
+
+This lives in its own module (not ``message_utils``, which is stdlib-only and is
+shipped inside the standalone viewer image) so the Telethon import never reaches
+the viewer.
+"""
+
+from __future__ import annotations
+
+from telethon.errors import LocationInvalidError, RPCError
+
+# Server message codes (already upper-case) handled by re-fetch + backoff retry.
+MEDIA_LOCATION_ERROR_MESSAGES = frozenset({"LOCATION_NOT_AVAILABLE", "LOCATION_INVALID"})
+
+
+def is_media_location_error(exc: BaseException) -> bool:
+    """Return ``True`` for a refreshable Telegram media-location error.
+
+    Matches precisely (no loose substring): by exact ``exc.message`` for codes
+    Telethon surfaces as a generic ``BadRequestError`` (e.g. ``LOCATION_NOT_AVAILABLE``)
+    and by exception type for codes with a generated class (``LocationInvalidError``).
+    """
+    if not isinstance(exc, RPCError):
+        return False
+    if isinstance(exc, LocationInvalidError):
+        return True
+    message = getattr(exc, "message", None)
+    return isinstance(message, str) and message.upper() in MEDIA_LOCATION_ERROR_MESSAGES
--- a/src/parallel_download.py
+++ b/src/parallel_download.py
@@ -246,6 +246,11 @@ class ParallelDownloader:
                 # The file actually lives on another DC; native download_media
                 # handles migration, so bail out to single-stream for this file.
                 raise ParallelDownloadUnavailable(f"file migrated to DC {exc.new_dc}") from exc
+            # Any other RPCError (incl. LOCATION_NOT_AVAILABLE / LOCATION_INVALID
+            # from upload.GetFile) intentionally propagates unchanged so the
+            # caller's refresh + retry loop can handle it. Do NOT add a broad
+            # `except Exception -> ParallelDownloadUnavailable` here: it would
+            # mask those as a stale single-stream fallback (see is_media_location_error).
             data = result.bytes
             if not data:
                 # Empty response before EOF would leave a gap; coverage check
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -101,6 +101,17 @@ def _media_retry_backoff_seconds(attempt: int) -> float:
     return base + random.uniform(0.5, 1.5)
 
 
+def _is_non_retryable_media_op(exc: BaseException) -> bool:
+    """Errors the media-download loop handles itself, so ``call_with_flood_retry``
+    must re-raise them rather than retry: location errors (the outer loop refreshes
+    and backs off) and a per-operation ``TimeoutError`` (the outer loop decides).
+
+    Keeping these out of ``call_with_flood_retry`` also means the per-operation
+    timeout never wraps — and so never cancels — its FloodWait sleeps.
+    """
+    return is_media_location_error(exc) or isinstance(exc, TimeoutError)
+
+
 def _pre_generate_thumbnail(source_path: str, media_root: str) -> None:
     """Pre-generate 200px WebP thumbnail for gallery grid view."""
     try:
@@ -1745,10 +1756,19 @@ class TelegramBackup:
         blows up the surrounding retry loop. Handles a deleted/unavailable
         message (``[]`` or ``[None]``) by returning ``None``.
         """
-        try:
-            fresh_messages = await asyncio.wait_for(
-                call_with_flood_retry(self.client.get_messages, chat_id, ids=[message.id]),
+
+        async def _get_messages_once():
+            # Time only the single Telegram call, so call_with_flood_retry still
+            # owns (and is never cancelled mid-) any FloodWait sleep.
+            return await asyncio.wait_for(
+                self.client.get_messages(chat_id, ids=[message.id]),
                 timeout=MEDIA_REFRESH_TIMEOUT_SECONDS,
+            )
+
+        try:
+            fresh_messages = await call_with_flood_retry(
+                _get_messages_once,
+                non_retryable=lambda exc: isinstance(exc, TimeoutError),
             )
         except (TimeoutError, RPCError, ConnectionError, OSError) as e:
             logger.debug("Could not refresh media reference (%s)", type(e).__name__)
@@ -1756,6 +1776,20 @@ class TelegramBackup:
         if fresh_messages and fresh_messages[0]:
             return fresh_messages[0]
         return None
+
+    async def _fetch_media_bytes_bounded(self, message: Message, tmp_path: str, file_size: int, timeout_val):
+        """``_fetch_media_bytes`` bounded by a per-operation timeout.
+
+        Timing only the single download operation (rather than the whole
+        ``call_with_flood_retry`` wrapper) ensures a Telegram FloodWait sleep is
+        never cancelled by the download timeout. A timed-out operation raises
+        ``TimeoutError``, which ``_is_non_retryable_media_op`` lets propagate to
+        the outer retry loop.
+        """
+        coro = self._fetch_media_bytes(message, tmp_path, file_size)
+        if timeout_val is None:
+            return await coro
+        return await asyncio.wait_for(coro, timeout=timeout_val)
 
     async def _download_media_to_path(self, message: Message, tmp_path: str, file_size: int, chat_id: int):
         """Download a message's media to ``tmp_path`` with bounded refresh + retry.
@@ -1780,15 +1814,13 @@ class TelegramBackup:
                 if os.path.exists(tmp_path):
                     os.remove(tmp_path)
                 try:
-                    return await asyncio.wait_for(
-                        call_with_flood_retry(
-                            self._fetch_media_bytes,
-                            message,
-                            tmp_path,
-                            file_size,
-                            non_retryable=is_media_location_error,
-                        ),
-                        timeout=timeout_val,
+                    return await call_with_flood_retry(
+                        self._fetch_media_bytes_bounded,
+                        message,
+                        tmp_path,
+                        file_size,
+                        timeout_val,
+                        non_retryable=_is_non_retryable_media_op,
                     )
                 except (FileReferenceExpiredError, RPCError) as e:
                     is_expired_ref = isinstance(e, FileReferenceExpiredError)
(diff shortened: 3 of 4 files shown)


GeiserX:
Hi @charys117 — thank you again for another well-crafted contribution. 🙏 The code is clean and idiomatic, the tests are thorough and symmetric (both the dedup and direct paths, the fail-open cases, the non-matching-RPC case), they're green, and you even followed our no-PII logging rule in the new log lines. Genuinely appreciated.

I ran a deep multi-perspective review on this one, and it surfaced one foundational question that I want to settle **before** we go further — so I've opened **#204** to track it properly (we like changes to start from a well-explained issue, and this is exactly the kind of thing that benefits from one).

## The core question: is `LOCATION_NOT_AVAILABLE` a real Telegram error?

I couldn't find evidence that Telegram actually emits this string, which would mean the new handler never fires in production:

- It's **not** in Telethon 1.43.2 (`telethon/errors/rpcerrorlist.py`). The closest real code is **`LOCATION_INVALID`** → `LocationInvalidError`.
- It's not in the public Telegram error databases (core.telegram.org/api/errors, the TL `errors.csv`, MTProto-Devs/RPC-errors, danog/rpc-db) or Pyrogram's catalogue.
- There's no traceback attached, and the string appears nowhere else in the codebase.

Because both the `non_retryable_rpc_messages` bypass and the outer `except RPCError` refresh are gated on that same string, as written the PR is effectively inert — nothing changes at runtime.

**Big caveat — I might be wrong.** Telegram sometimes returns server-side codes that aren't in any public list, so it's entirely possible you saw `LOCATION_NOT_AVAILABLE` verbatim in a real run. A traceback would confirm it instantly. This needs more digging on my side too, so please don't read this as a "no" — it's a "let's pin down the real trigger first."

**And there may well be a real gap here.** Telethon already auto-handles the related cases for us — `FileMigrateError` (DC switch, `client/downloads.py:112`), `FilerefUpgradeNeededError`/`FileReferenceExpiredError` (`:118`, refreshing `file_reference` at `:138`) — but `LocationInvalidError` is only caught in the *profile-photo* path (`:316`). On **regular message media** it would propagate uncaught to us. So your instinct (handle a location error that Telethon doesn't fix for us, by refreshing) could be right — most likely for `LOCATION_INVALID` rather than `LOCATION_NOT_AVAILABLE`. (One nuance: re-fetching the message refreshes `file_reference`, which is the documented fix for `FILE_REFERENCE_EXPIRED`; whether it actually helps a *location* error is unproven and is part of what #204 needs to confirm.)

### Could you help us pin it down?
Any one of these unblocks it:
1. **Share the real traceback / exact `RPCError` string** you observed (redact any chat/message IDs — we treat those as PII).
2. **Confirm whether you meant `LOCATION_INVALID`** (the real, documented code) — if so we can re-aim this PR at it.
3. If you'd rather, I'm happy to take it from here on #204 and pair with you on a targeted fix.

---

## Full review findings

For transparency, here's everything the review panel found — useful for whatever form the real fix takes. Note that the **High** design items below are currently *moot* (the branch never fires), but they'd become real the moment this is wired to a code that does, so they're worth baking into the fix.

<details><summary><b>High</b> — would be real bugs once the branch actually fires</summary>

- **`src/telegram_backup.py` (the new `non_retryable_rpc_messages` + outer loops)** — **Backoff is removed for this error.** Today: up to 5 retries with exponential backoff+jitter (≈2s→300s). Proposed: 3 back-to-back attempts with **no sleep**, plus 2 extra `get_messages` round-trips — a much higher instantaneous API rate per failing file, which risks a self-inflicted `FloodWait`/rate-limit. Any real fix should **keep exponential backoff** (and if the error is permanent server-side, prefer "give up cleanly" over retrying).
- **Worst case when refresh returns `None`** the file gets exactly **1** download attempt (was 6) before permanent failure — a resilience regression for transient errors.
</details>

<details><summary><b>Medium</b></summary>

- **Misleading log:** on the final attempt the code logs `"…refreshing before retry (attempt 3/3)"` and then immediately `raise`s, without refreshing or retrying. Emit it after the `if attempt == 2: raise` guard.
- **Pre-existing PII (worth fixing while here):** the `FileReferenceExpired` and timeout branches still log `message.id`. Your *new* lines correctly avoid it — bringing the old ones in line would close a small PII leak.
- **Over-engineered for one code:** `const → 1-element frozenset → generic _rpc_error_matches → _is_location_not_available_error` collapses to a single precise check (`isinstance(exc, RPCError) and exc.message == CODE`).
- **Misleading name:** `_is_location_not_available_error` actually tests membership in the whole set, so it'd misbehave the moment the set grows.
- **Policy conflation:** one frozenset drives two different decisions (skip-backoff *and* refresh-message) across two layers; they stay consistent only by coincidence.
- **Parallel downloader asymmetry** (`src/parallel_download.py`): `FileReferenceExpiredError` is explicitly re-raised through it, but a location error isn't — a future change to `_worker` could silently swallow it.
- **No distinct terminal signal** for "media permanently gone" vs a transient failure; the refresh-returns-`None` path fails silently into the generic handler.
- **Log-spam risk:** the new `info` lines fire in an unthrottled burst (no `FLOOD_WAIT_LOG_THRESHOLD`-style suppression).
- **`_rpc_error_matches` rebuilds the normalized set on every call** over static data — pre-normalize once.
- **Tests:** `test_rpc_error_matches_string_fallback` sets `.message=None` *after* construction, which doesn't represent a genuine unknown subclass (Telethon caches `str()` at construction). The **parallel-download path is untested** (`_should_parallelize` is always False under the MagicMock config), and there's no test for a `FloodWaitError` raised inside `_refresh_message_for_media`.
</details>

<details><summary><b>Low</b></summary>

- The refresh call runs **outside `asyncio.wait_for`**, so `download_timeout_seconds` doesn't bound it (worst case a long stall under flood backoff).
- A `FloodWaitError` during refresh is **silently absorbed** → the backup proceeds while the account is still flood-waited.
- **Substring fallback** (`code in str(exc)`) can over-classify / mask a different error whose text merely contains the substring; the exact `.message` check is the reliable one.
- The synthetic terminal `raise FileReferenceExpiredError(request=None)` discards the real last error (uninformative outer log).
- The `FileReferenceExpired` branch has no `attempt==2` guard (does a wasted 3rd refresh) while the new location branch does — minor inconsistency.
- `_refresh_message_for_media` doesn't pass `non_retryable_rpc_messages`, so the same code would get different retry handling in refresh vs download.
- An empty frozenset passed as `non_retryable_rpc_messages` silently no-ops the guard.
- Missing tests: `get_messages() == []` (vs `[None]`); the "2-fail-then-succeed" intermediate path; asserting the *exception type* in the final-attempt tests.
- Possible orphaned `.part` file on the new terminal `raise` (cleanup only runs at attempt start).
- The new path logs flat `info` and never escalates to `warning`/`error`, unlike the FloodWait convention.
</details>

<details><summary><b>Nits</b></summary>

- `LOCATION_NOT_AVAILABLE` constant is used exactly once.
- `test_rpc_error_matches_rpc_message` passes lowercase input to exercise case-insensitivity; a one-line comment would make that intent obvious.
- Interleaved concurrent retry logs are uncorrelatable (an inherent PII/observability tension — not really actionable).
</details>

<details><summary>Things we checked that are <b>fine</b> (not bugs)</summary>

- The trailing `raise FileReferenceExpiredError(request=None)` is **reachable, not dead** (the FRE-on-final-attempt-with-successful-refresh path reaches it).
- No `nonlocal message` corruption — it's a fresh closure per call and downloads are sequential.
- `_refresh_message_for_media` handles both `[]` and `[None]` safely (no `IndexError`).
- The new parameter adds zero overhead to the success path.
- Exception ordering (`FileReferenceExpiredError` before `RPCError`) is correct.
</details>

---

I'm going to hold off on merging until we've nailed down the real error in **#204** — I don't want to land a handler we can't prove will ever run, or one that trades away our backoff. Thank you again for the effort and the excellent test coverage; once we know the real code, a lot of this work carries straight over. 🚀

charys117:
I posted the production evidence and Telethon mapping details in #204: https://github.com/GeiserX/Telegram-Archive/issues/204#issuecomment-4812479806

Short version: I agree this is not a generated Telethon exception class, but Telethon preserves unknown 400 RPC strings as a generic `BadRequestError.message`. That matches the production log shape exactly: `RPCError 400: LOCATION_NOT_AVAILABLE (caused by GetFileRequest)`. I also found Telegram Desktop source handling `LOCATION_NOT_AVAILABLE` in `upload.GetFile` failures.

I agree the current PR should not be treated as automatically mergeable just from that evidence. Once we settle the trigger in #204, I can rework this PR to preserve backoff, use exact matching, bound refresh calls, avoid PII logging, and make terminal failure behavior explicit.

GeiserX:
Thanks @charys117 — your evidence settles it, and I owe you a correction: **you were right, `LOCATION_NOT_AVAILABLE` is a genuine error and my initial "not a real error" call was wrong.** I verified your explanation against the installed Telethon 1.43.2: `rpc_message_to_error()` falls back by numeric code (400 → `BadRequestError`) for unknown messages and preserves the raw string in `.message`, so the conversion reproduces your production log line exactly (`RPCError 400: LOCATION_NOT_AVAILABLE (caused by GetFileRequest)`). The Telegram Desktop reference you found corroborates it. Full write-up in #204.

To keep momentum, I pushed a reworked implementation on top of your branch that addresses the review while preserving your core idea (re-fetch the message to get a fresh reference/location). Happy to adjust anything — it's your PR and your reproduction.

### What changed vs. the original
- **New `src/media_errors.py`** classifies the error precisely: by `.message` for `LOCATION_NOT_AVAILABLE` (generic `BadRequestError`) and **by type** for `LocationInvalidError` (its `.message` is just `"BAD_REQUEST"`, so a text match would miss it). Kept out of `message_utils` so the Telethon import never reaches the viewer image.
- **Backoff preserved.** The official client treats this as a *transient* "file currently unavailable" (your case recovered only on a later run), so retries now back off with jitter instead of firing 3 sleepless attempts; an expired `file_reference` is fixed by the refresh itself and still retries immediately.
- **Fail clean.** After `MEDIA_REFRESH_MAX_ATTEMPTS` the item is left for the next scheduled backup run, and the *real* error propagates (no synthetic `FileReferenceExpiredError`). A refresh that returns nothing no longer gives up after a single attempt.
- **Both retry loops unified** into `_download_media_to_path` (dedup + direct), which also cleans up partial `.part` files on failure/cancellation.
- **Precise matching** (no loose substring); **bounded best-effort refresh** (timeout-guarded, swallows FloodWait so it can't hang or hammer); **no message IDs in logs** (also fixed the pre-existing ID logging in the adjacent branches).
- `call_with_flood_retry` now takes a `non_retryable` predicate instead of a message set; the parallel-download worker is documented so location errors keep propagating to the refresh loop.

### Tests
Reworked + expanded — including the real `rpc_message_to_error` wire conversion, the `LocationInvalidError`-by-type path, backoff-is-applied vs. expired-ref-skips-backoff, two-failures-then-success, and that the original error (not a synthetic one) propagates on exhaustion. Full suite: **1923 passed**; `ruff check` + `ruff format --check` clean.

### One ask before we merge
This only reproduces on your deployment. Could you pull this branch and confirm it actually recovers the affected media in a real run? That's the one thing tests can't prove. Once you've validated it, we'll bump the version, add a changelog credit to you, and release.

charys117:
Quick update: I agree that the short production smoke test only proved production compatibility / no obvious regression, not the actual `LOCATION_NOT_AVAILABLE` recovery path.

I have now redeployed the reworked PR #203 image for a longer real-world observation window.

Current test deployment:

- Deployment method: Flux/GitOps
- GitOps commit: `5ce085a`
- Flux applied revision: `main@sha1:5ce085a2de043aea3772a4f55b70574bd3b8ffa9`
- Backup image: `ghcr.io/charys117/telegram-archive:fork-c94a1351c11c`
- Runtime digest: `ghcr.io/charys117/telegram-archive@sha256:5890b95dc6aa885dfbc8b54297d7cc62bf1ce95d3a166247f65768f8db71f954`
- Viewer image remains official: `docker.io/drumsergio/telegram-archive-viewer:v7.17.1`

Baseline after rollout:

- Pod is `Running`, `2/2 Ready`
- Restart counts: `0`
- Scheduler started
- Backup started
- Listener started
- No Kubernetes warning events observed
- No `LOCATION_NOT_AVAILABLE`, `LOCATION_INVALID`, `RPCError 400`, `Traceback`, `ERROR`, or `maximum number of running instances reached` observed immediately after rollout

I have scheduled follow-up read-only checks at approximately 6h, 12h, 24h, 2d, and 3d from deployment.

If the transient location error occurs during this window, I will report whether it failed cleanly, whether the scheduler continued, and whether the media later recovered in a subsequent run. If it does not recur, I will report that honestly as extended production compatibility evidence, but I will not claim that the recovery branch was exercised.

GeiserX:
Thank you for the rigorous deployment writeup, @charys117 — pinning the GitOps commit, runtime digest, and a scheduled 6h/12h/24h/2d/3d observation window (and being explicit that a clean window is *compatibility* evidence, not proof the recovery branch fired) is exactly the kind of honest reporting we appreciate. 🙏

**We're going to merge this** — it's sufficiently solid: the trigger is confirmed real, the implementation is reviewed and reworked, CI is green, and your production run shows no regression. So this isn't blocked on the recovery path being observed.

That said, **please do keep reporting** as your window progresses:
- If `LOCATION_NOT_AVAILABLE` recurs: whether it failed cleanly, whether the scheduler kept running, and whether the media recovered on a later run (i.e. the recovery branch actually exercised).
- If it doesn't recur in 3 days: that's still useful as extended compatibility evidence — just report it as such.

Either way the data helps us tune `MEDIA_REFRESH_MAX_ATTEMPTS` / backoff if needed.

### Latest commit (`496906a`) — review follow-ups
- **FloodWait vs. timeout**: timeouts now wrap only the single Telegram operation (`_fetch_media_bytes_bounded` / a bounded `get_messages`), so `call_with_flood_retry` owns the rate-limit sleeps and a per-download/refresh timeout can no longer cancel a FloodWait mid-wait and fire the next request early. (Addresses the CodeRabbit "wait_for cancels FloodWait" finding.)
- **Clearer logs**: a failed refresh now logs "could not refresh … retrying anyway" instead of falsely claiming the reference was refreshed. (CodeRabbit finding.)
- **Coverage**: added tests for the backoff curve, the non-retryable predicate, per-operation timeout bounding, timeout retry/give-up, and partial-`.part` cleanup — patch coverage is now 100% on the changed lines (fixes the `codecov/patch` failure). Full suite: 1929 passing.

Thanks again — merging shortly.

PR #206, opened 2026-07-01
feat(messages): preserve previous message versions

charys117:
## Summary

- Preserve previous text versions for edited messages in a dedicated `message_versions` table.
- Add authenticated versions API/export support and a lazy-loaded viewer drawer with version counts.
- Keep reaction-only edit markers visible as `edited`, but only make `edited(n)` clickable when retained versions exist.

## Type of Change

- [ ] Documentation update

## Database Changes

- [x] Schema changes (Alembic migration required)
- [ ] No database changes

## Data Consistency Checklist

- [x] All `chat_id` values use marked format (via `_get_marked_id()`)
- [x] All datetime values pass through `_strip_tz()` before DB operations
- [x] INSERT and UPDATE operations handle the same fields identically

## Testing

Targeted local validation:

- `uv run --extra dev pytest tests/test_db_adapter_soft_delete_upsert.py tests/test_web_routes.py tests/test_web_coverage.py tests/test_frontend_bootstrap.py -q`
- `uv run --extra dev pytest tests/test_db_adapter.py::TestDeleteChatOperations::test_delete_chat_issues_six_deletes -q`
- `uv run --extra dev ruff check src/db/adapter.py src/web/main.py src/db/migrate.py tests/test_db_adapter_soft_delete_upsert.py tests/test_frontend_bootstrap.py tests/test_web_routes.py tests/test_web_coverage.py`
- `uv run --extra dev ruff check tests/test_db_adapter.py`
- `git diff --check`

CI/fork validation before the latest review-cleanup amend:

- Full upstream PR CI passed for commit `2878206`.
- Fork GHCR test images built successfully: https://github.com/charys117/Telegram-Archive/actions/runs/28488550242

## Security Checklist

## Deployment Notes

- Alembic migration `015` creates the `message_versions` table.
- Pre-Alembic schema stamping detects `message_versions` for both PostgreSQL and SQLite.
- Docker image rebuild required for deployment.
- This feature branch was deployed and observed running stably before opening the PR.

## Summary by CodeRabbit

* **New Features**
  * Added preserved message edit version history across exports and the web viewer, including a new versions endpoint and a “Message Versions” drawer with live update support.
* **Bug Fixes**
  * Improved edit processing to be edit-date–aware and idempotent (no-ops won’t create extra versions), with correct timezone handling.
  * Ensured deletions remove version history safely and that the UI shows accurate per-message version counts.
* **Chores**
  * Enhanced schema detection/stamping and migration verification for missing legacy tables, plus more consistent migration dry-run counting.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -9,6 +9,9 @@ on:
     paths:
       - 'Dockerfile'
       - 'requirements.txt'
+      - 'pyproject.toml'
+      - 'uv.lock'
+      - 'scripts/**'
       - 'src/**'
       - '.github/workflows/docker-publish.yml'
       # Exclude viewer-only changes
--- a/alembic/versions/20260626_015_add_message_versions.py
+++ b/alembic/versions/20260626_015_add_message_versions.py
@@ -1,0 +1,77 @@
+"""Add message versions table.
+
+Revision ID: 015
+Revises: 014
+Create Date: 2026-06-26
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "015"
+down_revision: str | None = "014"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+TABLE_NAME = "message_versions"
+IDX_MESSAGE_DATE = "idx_message_versions_message_date"
+IDX_MESSAGE_CAPTURED = "idx_message_versions_message_captured"
+
+
+def _table_exists(inspector: sa.Inspector) -> bool:
+    return TABLE_NAME in inspector.get_table_names()
+
+
+def _index_exists(inspector: sa.Inspector, name: str) -> bool:
+    return name in {idx["name"] for idx in inspector.get_indexes(TABLE_NAME)}
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    inspector = sa.inspect(conn)
+
+    if not _table_exists(inspector):
+        op.create_table(
+            TABLE_NAME,
+            sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+            sa.Column("message_id", sa.BigInteger(), nullable=False),
+            sa.Column("chat_id", sa.BigInteger(), nullable=False),
+            sa.Column("text", sa.Text(), nullable=True),
+            sa.Column("date", sa.DateTime(), nullable=False),
+            sa.Column("change_hash", sa.String(length=64), nullable=False),
+            sa.Column("captured_at", sa.DateTime(), server_default=sa.func.now(), nullable=False),
+            sa.ForeignKeyConstraint(
+                ["message_id", "chat_id"],
+                ["messages.id", "messages.chat_id"],
+                name="fk_message_versions_message",
+                ondelete="CASCADE",
+            ),
+            sa.PrimaryKeyConstraint("id"),
+            sa.UniqueConstraint("change_hash", name="uq_message_versions_change_hash"),
+        )
+        inspector = sa.inspect(conn)
+
+    if not _index_exists(inspector, IDX_MESSAGE_DATE):
+        op.create_index(IDX_MESSAGE_DATE, TABLE_NAME, ["chat_id", "message_id", "date"])
+        inspector = sa.inspect(conn)
+    if not _index_exists(inspector, IDX_MESSAGE_CAPTURED):
+        op.create_index(IDX_MESSAGE_CAPTURED, TABLE_NAME, ["chat_id", "message_id", "captured_at"])
+
+
+def downgrade() -> None:
+    conn = op.get_bind()
+    inspector = sa.inspect(conn)
+
+    if not _table_exists(inspector):
+        return
+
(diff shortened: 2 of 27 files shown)


GeiserX:
Thank you @charys117 — this is an impressively complete feature: the version model is semantically right (reaction-only edits bump `edit_date` without fake versions, imports can't clobber newer live data), migration 015 passed a full executed idempotency matrix including partial states, the entrypoint stamping is correct in both blocks, the new endpoint's auth/IDOR posture is solid, and the standalone-viewer import chain is verified clean. You also quietly **fixed a real pre-existing bug** (re-backups resetting `is_pinned`) and removed a PII log. And thanks for addressing the earlier CodeRabbit round before we even looked. 🙏

I ran a 13-perspective review panel over it (correctness, architecture, security, concurrency, performance — with benchmarks — data integrity, Telegram domain semantics, API compatibility, error handling, tests, DevOps, UX/a11y, simplicity). **No critical defects.** As with #203, I've applied the panel's findings directly on your branch (`331b0b4`) so we don't lose momentum — happy to adjust anything you disagree with. Full suite: **1977 passed**; ruff clean.

## What the review found → what changed

**Performance (the big one).** The reworked upsert paid 4–5 statements per *already-archived* message (measured ~56× slower on the conflict path; a 5M-message full re-backup gained ~12–15 min, and every unchanged row took the SQLite write lock). Now: a lock-free snapshot check skips unchanged messages entirely, and updates write only the columns that actually differ. Incremental runs were always fine; full re-backups now are too.

**Robustness.** `_record_message_version` runs inside a SAVEPOINT, so an unexpected version-insert failure can no longer roll back a whole message batch. `insert_message` and `update_message_text` gained `@retry_on_locked` (a transient SQLite lock could silently drop a listener edit). The conflict-then-vanished-row race now logs (PII-free) instead of dropping silently.

**Honest counters + realtime consistency.** `update_message_text` returns `applied`/`noop`/`not_found`; the listener counts skips separately and — importantly — only broadcasts the websocket `edit` event when the archive actually changed, so the viewer can no longer display text the archive rejected as stale, and `edited(n)` counts can't drift.

**A pre-existing sync bug your feature made heavier.** `_sync_deletions_and_edits` compared `str(remote_edit_date)` (tz-aware) with the archived naive datetime — they *never* match, so every previously-edited message triggered a spurious update call on every sync pass (each now costing the locked path). Fixed with proper tz-normalized datetime comparison + a regression test.

**Export.** `message_versions` now streams entry-by-entry like messages (it was materializing a chat's entire edit history into one list + one `json.dumps` mid-stream — memory spike + a "200 then truncated JSON" failure window), and the versions query dropped its unnecessary join to `messages`.

⚠️ **One honest disagreement with the PR description:** the export change (top-level array → `{chat, messages, message_versions}`) **is a breaking change** to `GET /api/chats/{id}/export`, even though the shape itself is better. Blast radius is low (the viewer only downloads the file; no in-repo consumer parses it), so we're keeping your shape — but it will be flagged as a **breaking change in the CHANGELOG** at release rather than labeled non-breaking.

**Simplicity.** The `source` parameter was threaded through five signatures and seven call sites but never read anywhere — removed (if you intended it as a future audit hook, let's add it back *with* a consumer, e.g. a column or structured log). Also removed: the viewer's client-side re-sort of already-server-sorted versions.

**Migration/DevOps.** `_is_missing_table_error` now matches anchored dialect messages only (loose `"does not exist"` containment could classify real errors as "0 records" during migration verification). `docker-publish.yml` now also triggers on `scripts/**` (your entrypoint stamping change must always ship with the image), plus `pyproject.toml`/`uv.lock`. `captured_at`'s Python-side default no longer uses deprecated `datetime.utcnow`. A comment now documents that the explicit version deletes are load-bearing (SQLite runs with FKs off, so the CASCADE is inert there).

**Viewer polish.** Escape closes the drawer (matching the lightbox pattern), `role="dialog"`/`aria-modal` + initial focus, `aria-live` on the loading state, 401 routes to re-auth instead of a futile Retry loop, 503 gets a specific message, the truncation hint no longer false-positives at exactly 100, and version dates are labeled "as of …" (they mark when that text *became* current).

**Tests.** Added: `is_pinned` preservation (your fix deserved a guard), the documented refusal case (differing text with no edit evidence on either side), outcome statuses, savepoint containment, unchanged-reupsert no-op, migration partial-state **index recovery** assertions (the riskiest real-world state was unasserted), endpoint generic-500, and the aware-vs-naive sync regression. Removed: the vestigial `mock_session.get` setups (nothing calls `session.get`) and the most brittle string asserts (exact Tailwind classes, `"Before"/"After"` absence checks).

**Documented rather than changed** (deliberate calls, now written down in code comments): the `>` vs `>=` asymmetry between the upsert and live-edit policies; non-text metadata staying last-writer-wins when text is withheld; the change_hash encoding being a frozen contract (and its accepted caption-oscillation dedup edge); text-only versioning (formatting/entity-only edits aren't captured).

## Deployment note (for the release)
Once a DB is stamped `015`, rolling back to a ≤7.17.2 image crash-loops on `Can't locate revision '015'` — same as every migration here. The release notes will carry the escape hatch (`alembic downgrade 014` before downgrading the image).

Really nice work on this one — the hard parts (conflict policy, migration hygiene, lazy drawer with a stale-response guard) were already right. 🚀

PR #214, opened 2026-07-14
fix(viewer): stop realtime poll clobbering a jump-to-message window (#213)

### LINKED ISSUE #213: [Bug]: Jumping to a message from the media gallery snaps back to the latest messages
### Version

7.20.0

### Deployment Method

Docker Compose

### Database Type

PostgreSQL

### Bug Description

Clicking a media (e.g. a voice recording) item in the per-chat media gallery loads the correct history window from the server, and then the view is immediately reset to the newest messages. The gallery closes and the target message is never shown.

Both API requests return 200 — the backend behaves correctly. The frontend fetches the jump window (before_id=NNN) and then something immediately re-fetches from the top (offset=0), clobbering it.

Note: calendar jump-to-date IS fixed in 7.19.0 (#207) and works correctly on 7.20.0. This looks like the same class of bug on a path #207 did not cover — jumping OUT of the gallery INTO history, rather than the gallery-unmount case that was fixed. Suspect a stale poller or scroll observer surviving the gallery → history transition.

### Expected Behavior

The chat history should scroll to and highlight the message containing the selected media item.

### Steps to Reproduce

1. Open a chat with media in the web viewer.
2. Open the media gallery.
3. Select the Voice/Audio tab (also reproduces with other media types).
4. Click any item.
5. The gallery closes and the view returns to the newest messages instead of jumping to the message.

### Logs

```shell
INFO: "GET /api/chats/<chat_id>/media?types=voice%2Caudio&limit=50 HTTP/1.1" 200 OK
INFO: "GET /api/chats/<chat_id>/messages?before_id=217230&limit=50 HTTP/1.1" 200 OK   <-- jump window loads correctly
INFO: "GET /api/chats/<chat_id>/messages?limit=50&offset=0 HTTP/1.1" 200 OK           <-- immediately resets to top
```

### Environment

_No response_

### Checklist

GeiserX:
## Summary

Fixes #213. On v7.20.0, clicking a media item in the per-chat gallery to jump to its message loads the correct `before_id` history window (200), then an `offset=0` fetch immediately clobbers it and the view snaps back to the newest messages — the target is never shown.

## Root cause

`loadMessagesAroundId` **replaces** the message list with a detached window around the target (`messages.value = data`) that does **not** contain the live tail. The 3s realtime poll (`checkForNewMessages`, `?limit=50&offset=0`) then fetches the latest page and `upsertMessages` adds it; because the jumps smooth-scroll is still animating from the bottom, `shouldStickToBottom` reads true and the poll calls `scrollToBottom()` → snap to newest.

Calendar **jump-to-date** was unaffected (as #213 notes) because `jumpToDate` *keeps* the existing tail in the list (`messages.value.push(message)` + background gap-fill), so the polls upsert of the latest page is a no-op.

## Fix

A `viewingPinnedWindow` flag: `loadMessagesAroundId` sets it (after its `resetMessagePagination`), `checkForNewMessages` bails while it is set, and `resetMessagePagination` clears it — so the poll resumes automatically the moment any `openChat`/`openTopic`/`search` re-establishes a tail-inclusive view. In-flight polls started before the jump are already invalidated by the existing `++chatVersion` guard, so both the racing and the future poll are covered. Realtime updates pause only while you are reading a detached jump window, which is the correct behavior.

## Testing

- New `test_jump_window_suppresses_realtime_poll` pins the flag, the poll guard, the set-after-reset ordering, and the clear in `resetMessagePagination`.
- Full suite **2026 passed / 0 failed**; `ruff` clean; viewer JS `node --check` clean.

No release cut per request.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented realtime message refresh from disrupting “jump-to-message” browsing in a detached message window.
  * Restored live, tail-inclusive realtime updates when returning to the standard latest view.
* **Tests**
  * Added regression coverage to ensure realtime polling is suppressed during pinned/jump windows and re-enabled when exiting.
  * Updated realtime-search skip and unseen-message related tests to match recent UI/template behavior.

### DIFF
--- a/src/web/templates/index.html
+++ b/src/web/templates/index.html
@@ -1742,6 +1742,12 @@
                 // monotonically older).
                 let oldestMessageCursor = null
                 let loadFailureStreak = 0
+                // True while the view is a detached jump-to-message window (loadMessagesAroundId)
+                // that does NOT contain the live tail. The realtime poll fetches the latest page
+                // (offset=0) and would upsert it in and snap the view back to newest, so the poll
+                // is suppressed until a tail-inclusive view is re-established (via
+                // resetMessagePagination, which every openChat/openTopic/search entry calls).
+                let viewingPinnedWindow = false
 
                 const messageIdKey = (msg) => msg?.id == null ? null : String(msg.id)
 
@@ -1784,6 +1790,10 @@
                     hasMore.value = true
                     oldestMessageCursor = null
                     loadFailureStreak = 0
+                    // Every openChat/openTopic/search entry loads a tail-inclusive view through
+                    // here, so clear the pinned-window flag (loadMessagesAroundId re-sets it after
+                    // its own reset). This also re-enables the realtime poll.
+                    viewingPinnedWindow = false
                     // View-entry chokepoint: every message-list swap goes through here, so the
                     // unseen-badge from the previous view resets alongside pagination.
                     unseenMessageCount.value = 0
@@ -3072,6 +3082,10 @@
                             if (chatVersion !== myVersion) return
                             messages.value = data.messages || data
                             resetMessagePagination()
+                            // This window is detached from the live tail — pause the realtime
+                            // poll so it can't fetch the latest page and snap us back to newest
+                            // (#213). Cleared when a tail-inclusive view is re-established.
+                            viewingPinnedWindow = true
                             await nextTick()
                             if (chatVersion !== myVersion) return
                             setupMessagesScrollObserver()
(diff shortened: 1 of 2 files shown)


GeiserX:
Re: CodeRabbit's unittest.TestCase suggestion — skipping intentionally. `tests/test_frontend_bootstrap.py` is, by design, entirely module-level pytest functions with bare asserts (it does static string-assertions against the built `index.html`, no mocks/fixtures); there is no TestCase in the file. The new tests match that established local convention exactly — wrapping just these in a TestCase would make them the sole inconsistent ones. The repo's general "use unittest.TestCase" guideline applies to the mock-based suites (adapter/backup/listener), not this static-HTML file.

PR #215, opened 2026-07-14
fix(media): budget filenames for eCryptfs + cap failed-download retries (#212)

### LINKED ISSUE #212: [Bug]: Media silently lost on Synology encrypted shares — filenames exceed the ~143-char eCryptfs limit
### Version

v7.20.0

### Deployment Method

Docker Compose

### Database Type

PostgreSQL

### Bug Description

Media whose original filename is long cannot be written to a Synology **encrypted** shared folder, and is silently skipped. The message itself is archived; only the media file is missing.

Non-Latin scripts hit this much sooner: Cyrillic/CJK are 2+ bytes per character in UTF-8.

Root cause: Synology encrypted shares use eCryptfs, whose filename-encryption overhead reduces the usable filename length to roughly 143 characters — not the normal 255-byte filesystem limit.

The download path writes to a temp file before renaming:

    <media_id>_<original_name>.mp4.1.<id>.part

so the temp name is LONGER than the final name. A file whose final name would fit can still fail at the .part stage. The app makes no attempt to fit the name within the filesystem's limit.

Impact: the media row is created with downloaded=0 and file_path NULL, so it is "tracked but missing". _retry_pending_media_downloads() then re-attempts it on EVERY subsequent scheduled run, forever — it can never succeed while the name is too long. This is a permanent, silent tax on every backup run, and permanent data loss for the affected files.

Observed: 6 media files lost out of ~19,900, in a 490-chat / 167k-message archive, all with long Cyrillic filenames.

### Expected Behavior

The filename component should be truncated to fit the filesystem's limit before writing.

The <media_id> prefix is already unique, so the original filename is decorative and can safely be shortened or dropped entirely.

Two details that matter for the fix:
- The limit must be measured in BYTES, not characters (UTF-8 multibyte).
- The temp suffix (.<n>.<id>.part) must be included in the length budget, since it is what actually overflows.

### Steps to Reproduce

1. Create an ENCRYPTED shared folder on a Synology NAS.
2. Mount it on the Docker host via CIFS and point BACKUP_PATH at it.
3. Back up a chat containing media with long original filenames (e.g. Russian-language video titles).
4. Observe the failures in the pending-media retry phase.

Minimal filesystem-level reproduction (no app involvement needed):

    $ cd /mnt/nas/<encrypted-share>/media/<chat_id>
    $ touch "$(python3 -c "print('я'*120 + '.mp4.1.124206728753904.part')")"
    touch: cannot touch 'яяя…я.mp4.1.124206728753904.part': File name too long

120 Cyrillic chars = 240 bytes, plus a 26-byte temp suffix. Well under 255 characters — but over the ~143-char eCryptfs ceiling.

### Logs

```shell
Retrying 6 pending media downloads...
WARNING - Transient Error (FileNotFoundError): sleeping 2.59s before retrying _fetch_media_bytes (retry=1/5): [Errno 2] No such file or directory: '/data/backups/media/<chat_id>/<media_id>_<long cyrillic filename>.mp4.1.124206728753904.part'
WARNING - Transient Error (FileNotFoundError): sleeping 5.48s before retrying _fetch_media_bytes (retry=2/5): [Errno 2] No such file or directory: ...
WARNING - Transient Error (FileNotFoundError): sleeping 8.98s before retrying _fetch_media_bytes (retry=3/5): ...
WARNING - Transient Error (FileNotFoundError): sleeping 17.15s before retrying _fetch_media_bytes (retry=4/5): ...
WARNING - Transient Error (FileNotFoundError): sleeping 33.29s before retrying _fetch_media_bytes (retry=5/5): ...
```

### Environment

_No response_

### Checklist

GeiserX:
## Summary

Fixes #212. Media with long filenames was silently lost on Synology **encrypted** (eCryptfs) shares — whose filename-encryption overhead caps a name component at **~143 bytes**, not 255 — and the failed rows were then re-fetched from Telegram on **every** scheduled run forever. Non-Latin (Cyrillic/CJK) filenames hit the limit far sooner. Implements the studied **Option D** from the research (deterministic conservative byte-budget truncation, NOT pathconf-measured or ENAMETOOLONG-probing) plus the retry cap, per the two decisions taken on the issue.

## Why not the naive fix

A 6-agent research panel established that `os.pathconf`/`statvfs` reports **255** over CIFS→Synology (the check runs client-side, on the wrong side of the encryption — confirmed via eCryptfs Launchpad #885744 + the CIFS topology), so "measure the filesystem limit and fit" under-truncates and *is* the bug. Adaptive `ENAMETOOLONG`-probe was also rejected: the pipeline **recomputes** the filename from the message every run and relies on it matching what is on disk, so a probe-length-dependent name breaks dedup/retry. Hence a deterministic fixed budget + hash fallback.

## Filename budgeting

- New pure `build_media_filename()` (`message_utils.py`): byte-aware, extension-preserving, UTF-8 **codepoint-safe**, **deterministic** truncation of the decorative original-name stem, reserving the temp `.part` suffix. The `file_id` prefix (uniqueness) and extension are always kept; a short deterministic `sha1` fallback covers pathological tiny budgets. Called by **both** `_get_media_filename` sites (backup + listener), so the temp file, the sharded store leaf, and the chat symlink all inherit a length-safe name.
- `MEDIA_MAX_FILENAME_BYTES` (default **143**) makes it configurable; only long decorative names are shortened — short names are untouched.

## Retry cap (the "permanent silent tax")

- New `media.download_attempts` column — **Alembic 016** (idempotent; `entrypoint.sh` stamping updated for PG **and** SQLite) plus `MEDIA_MAX_DOWNLOAD_ATTEMPTS` (default **5**). `get_pending_media_downloads` excludes maxed-out rows; `_retry_pending_media_downloads` increments the counter on each unsuccessful re-attempt, so a permanently-unwritable file stops being re-fetched.

## Testing

- Pure unit tests for the truncation helper (byte budget, extension preserved, multibyte-safe, deterministic, hash fallback).
- **Real-SQLite** migration-016 tests (idempotent up/down, server_default) and retry-cap adapter tests (cap filter + increment).
- Retry-loop increment-wiring tests.
- Full suite **2049 passed / 0 failed**; `ruff` clean.

No release cut per request (batching with #213).

## Summary by CodeRabbit

* **New Features**
  * Added configurable limits for media filename length and download retries.
  * Media filenames now remain safe, deterministic, and within configured byte limits.
  * Failed media downloads are tracked and automatically skipped after reaching the retry limit.
  * Redownload requests reset the retry counter.

* **Bug Fixes**
  * Improved migration handling for databases upgrading to the latest media schema.
  * Added warnings when media downloads are permanently skipped after repeated failures.

* **Tests**
  * Added coverage for filename limits, retry handling, and database migrations.

### DIFF
--- a/alembic/versions/20260714_016_add_media_download_attempts.py
+++ b/alembic/versions/20260714_016_add_media_download_attempts.py
@@ -1,0 +1,45 @@
+"""Add media.download_attempts (failed-download retry cap).
+
+Revision ID: 016
+Revises: 015
+Create Date: 2026-07-14
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "016"
+down_revision: str | None = "015"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+TABLE_NAME = "media"
+COLUMN_NAME = "download_attempts"
+
+
+def _column_exists(inspector: sa.Inspector) -> bool:
+    return COLUMN_NAME in {c["name"] for c in inspector.get_columns(TABLE_NAME)}
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    inspector = sa.inspect(conn)
+
+    # Idempotent: create_all()-provisioned databases may already have the column.
+    if TABLE_NAME in inspector.get_table_names() and not _column_exists(inspector):
+        op.add_column(
+            TABLE_NAME,
+            sa.Column("download_attempts", sa.Integer(), nullable=False, server_default="0"),
+        )
+
+
+def downgrade() -> None:
+    conn = op.get_bind()
+    inspector = sa.inspect(conn)
+
+    if TABLE_NAME in inspector.get_table_names() and _column_exists(inspector):
+        op.drop_column(TABLE_NAME, COLUMN_NAME)
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -121,6 +121,15 @@ if has_tables and not has_alembic:
     \"\"\")
     has_015_message_versions = cur.fetchone()[0]
 
+    # Check artifact from migration 016: media.download_attempts column
+    cur.execute(\"\"\"
+        SELECT EXISTS (
+            SELECT FROM information_schema.columns
+            WHERE table_name = 'media' AND column_name = 'download_attempts'
+        );
+    \"\"\")
+    has_016_download_attempts = cur.fetchone()[0]
+
     # Check artifact from migration 014: messages soft-delete marker columns
     cur.execute(\"\"\"
         SELECT
@@ -241,7 +250,9 @@ if has_tables and not has_alembic:
     has_push_subs = cur.fetchone()[0]
 
     # Determine which version to stamp based on existing schema
-    if has_015_message_versions and has_014_soft_delete:
+    if has_016_download_attempts and has_015_message_versions and has_014_soft_delete:
+        stamp_version = '016'
+    elif has_015_message_versions and has_014_soft_delete:
         stamp_version = '015'
     elif has_014_soft_delete:
         stamp_version = '014'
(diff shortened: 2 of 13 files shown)


GeiserX:
Re: CodeRabbit type-hint suggestion on the new test files — skipping intentionally for local consistency. The sibling test files these mirror carry no signature hints: test_migration_016.py matches test_migration_014/015.py, and test_media_retry_cap.py matches test_folder_resolution_integration.py / test_media_adapter.py (all use bare def test_...(): with untyped fixtures/helpers). Annotating only the two new files would make them the sole hinted migration/integration tests, inconsistent with the ones right beside them. The repo guideline to use type hints applies to production signatures in src/, which this PR does follow.

PR #216, opened 2026-07-14
fix(web): initialize media gallery state before use

cergmin:
## Summary

Fix of the client errors:

```
vue.global.prod.js:5 ReferenceError: Cannot access 'showMediaGallery' before initialization

vue.global.prod.js:5 TypeError: Cannot read properties of undefined (reading 'username')
```

- Move the Media Gallery state into the initial Vue `setup()` state block.
- Ensure `showMediaGallery` is initialized before callbacks and watchers reference it, preventing the page from failing during startup.

## Type of Change

- [ ] Documentation update

## Database Changes

- [x] No database changes

## Data Consistency Checklist

Not applicable — no database code was changed.

## Testing

## Security Checklist

- [x] User input handling is not affected
- [x] Authentication/authorization behavior is not affected

## Deployment Notes

- Rebuild the viewer Docker image to include the updated `index.html`.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized the Shared Media state declarations for improved code structure.
  * No user-visible behavior or functionality changed.

### DIFF
--- a/src/web/templates/index.html
+++ b/src/web/templates/index.html
@@ -1646,6 +1646,21 @@
                 const selectedDate = ref(null)
                 let flatpickrInstance = null
                 const showScrollToBottom = ref(false)
+
+                // Media Gallery state (v7.10.0)
+                const showMediaGallery = ref(false)
+                const mediaGalleryTab = ref('photos')
+                const mediaGalleryItems = ref([])
+                const mediaGalleryLoading = ref(false)
+                const mediaGalleryHasMore = ref(true)
+                const mediaGalleryCounts = ref({})
+
+                const mediaTabs = [
+                    { id: 'photos', label: 'Photos & Videos' },
+                    { id: 'voice', label: 'Voice' },
+                    { id: 'files', label: 'Files' },
+                ]
+
                 // Messages that arrived (WS/poll) while the user was scrolled up reading
                 // history — shown as a badge on the jump-to-latest button; cleared when the
                 // user returns near the bottom or switches view.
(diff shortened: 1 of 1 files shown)


GeiserX:
Thanks for this, @cergmin — and for paying attention to setup-time ordering, that's exactly the kind of bug that's miserable to chase.

I dug into it carefully and I can't reproduce either error on the current code, and I wanted to share the details in case it helps narrow down what you saw.

**On the ordering:** in `setup()` the state block containing `const showMediaGallery = ref(false)` already sits *before* the only two statements that evaluate `showMediaGallery` during setup — the `watch(showMediaGallery, …)` call and the final `return { … }`. This holds both on this PR's base (v7.21.0: declaration at line 2926, watcher at 3112) and on current `main` (v7.22.0: 2957 vs 3181). Every other reference to it is either in the template (evaluated only after `setup()` returns) or inside a function body like `selectChat`/`selectTopic` (runs only when called). Since Vue runs `setup()` top-to-bottom, the temporal-dead-zone window is never entered, so `Cannot access 'showMediaGallery' before initialization` can't fire from this source tree.

There's also a regression test that pins this invariant so it can't silently regress: `tests/test_frontend_bootstrap.py::test_media_gallery_refs_are_initialized_before_watcher` (it asserts the declaration precedes the watcher — and it was already present on v7.21.0).

**On the `username` TypeError:** there's no `currentUser.value.username`-style object access in the file (the binding is a flat `currentUsername` string ref), so that one only makes sense as a *cascade* of the first error — if `setup()` threw early, the template would render against undefined bindings. With the first error not firing, this one doesn't either.

My best guess is that what you hit came from a **stale or cached bundle** (the CDN `vue.global.prod.js` + an older `index.html`), a locally-modified file, or an earlier build where the watcher did precede the declaration — all plausible, just not present in the committed source at your base or on latest.

I'm going to leave this open rather than close it: if you can still reproduce on the latest `main` (or the `v7.22.0` release) with a hard refresh / cache cleared, please drop the exact steps and I'll happily dig back in. Note that `main` has moved a fair bit since you branched (the media-gallery/jump-to-message area was reworked), so a fresh pull may be worth a try first.

Really do appreciate the contribution 🙏

GeiserX:
Thanks @cergmin for catching this — clean fix, the gallery state was indeed declared after its first use. Merged!

PR #218, opened 2026-07-14
chore: whole-repo audit hardening (security, reliability, perf, PII)

GeiserX:
## Summary

Whole-repo health audit (`/review-code!`) run by a panel of specialized reviewers, with every actionable finding fixed here. **Stacked on #217** — review that first; this branch is based on it, so the diff below is audit work only. GitHub will retarget this PR to `main` automatically when #217 merges.

Full suite: **2161 passed**, `ruff check` + `ruff format --check` clean across the repo. An independent verifier pass reviewed the combined diff; the one blocker it found (an SSRF-filter encoding bypass) is fixed and re-verified in this branch.

## What was fixed

**Security**
- `ALLOW_ANONYMOUS_VIEWER` no longer grants **master/admin** to anonymous requests — it now yields a read-only viewer. Previously any anonymous request could create viewers, mint share tokens, read the audit log, and change settings.
- Push-subscribe endpoint (`/api/push/subscribe`) validates the target by **resolving the host and rejecting private/loopback/link-local/reserved addresses**, off the event loop. The prior fix used string matching, which was bypassable via decimal/hex/octal/shorthand IP encodings and trailing dots (blind SSRF into internal networks / cloud metadata).
- Chat-search `ILIKE` wildcards (`%`/`_`/`\`) are now escaped, matching the message-search path.

**Reliability**
- Realtime Postgres LISTEN connection is closed on every exit path (previously leaked one connection per retry while the DB flapped); callback tasks are tracked so their exceptions surface.
- Scheduler watchdog retries a listener that failed to start (previously one failed restart permanently disabled realtime capture until a container restart).

**Data correctness**
- `download_date` was written in host-local time by the backup path while the listener/import paths wrote UTC — now uniformly `utcnow_naive()`.
- Migrations **002–006** are now idempotent (inspector-guarded), matching the repo's mandate; unguarded creates could crash-loop `create_all()`-provisioned databases.

**Performance**
- The messages page endpoint issued up to ~100 sequential queries per 50-row page (per-message reactions + reply-text); now **4 queries per page** via batched `IN` lookups. Since the viewer polls this endpoint every 3s per open tab, this was continuous load.
- Per-chat stats gain a 60s TTL cache (ACL still enforced before any cache hit).
- `get_pending_media_downloads` is bounded (was materializing every pending row — a prior OOM class).

**Observability / PII**
- Scrubbed chat ids, topic ids, message ids, chat/topic titles, message text, and file paths from logs across listener/backup/import/realtime/push (the repo's own logging policy) — including one log line that interpolated verbatim message text and another that logged the raw notification payload.
- `_is_db_connection_error` now maps SQLite "database is locked" and asyncpg connection errors to 503 (were 500). Silent `json.loads`/`get_entity` bare-excepts narrowed and logged.

**Consistency / hygiene**
- Single shared `fallback_media_filename()` so the backup and listener ingest paths produce identical names for media with no usable original name (the #212 divergence class, moved into the fallback tier).
- Removed shipped frontend debug logs; documented the dual `chatVersion`/`requestSeq` staleness idioms.
- `models.py` datetime defaults on `utcnow_naive`; export timestamp in UTC; CI runs the suite once (was twice).

**Docs**
- Documented `MEDIA_MAX_FILENAME_BYTES` and `MEDIA_MAX_DOWNLOAD_ATTEMPTS`; added a loud warning that the reverse proxy MUST strip `AUTH_PROXY_HEADER` inbound (else header-forged auth bypass); documented `ALLOW_ANONYMOUS_VIEWER` as read-only.

## New test coverage (real, not shape-only)
- `test_jump_window_backend.py`, `test_messages_page_batching.py` (asserts the ≤6-query count via an engine event listener), `test_migrations_002_006_idempotency.py`, `test_migration_017.py`, `test_websocket_route.py` (real `websocket_connect`, incl. ACL-denial), and a rewritten `test_entrypoint_schema_detection.py` that **executes** the shipped SQLite stamping heredoc against seeded DBs (with a mutation check proving it isn't tautological). Frontend adds a generic drift guard: every `data-*` attribute queried from JS must actually be rendered in the template.

## Deliberately deferred (recommendations for maintainer decision — not in this PR)
Structural refactors (`backup_all`/listener decomposition, `DatabaseAdapter` split, `index.html` composables), a Playwright browser test (new dependency + CI browser install), vendoring the 5 CDN assets + tightening CSP, an FTS search index, a Postgres service in CI, API list-response-shape standardization, and send-time IP pinning for the push path (defense-in-depth beyond the resolution check here). The dead-code audit's "unused symbol" findings were **refuted during verification** (all had live call sites) — nothing was deleted on that basis.

## Summary by CodeRabbit

- **New Features**
  - Added configurable media filename byte limits and per-file download retry limits.
  - Added toast-based UI notifications for common background failures.
  - Added cached per-chat statistics and `after_id`/`before_id` jump-to-message windows.
  - Added proxy authentication options for safer deployments.

- **Bug Fixes**
  - Made anonymous viewing strictly read-only (no master/admin privileges).
  - Hardened search wildcard handling and improved message pagination consistency.
  - Improved realtime reliability during database blips; DB-unavailable now returns 503.

- **Documentation**
  - Clarified anonymous access and reverse-proxy header security requirements, plus environment variables.

- **Chores/Tests**
  - Expanded coverage and added regression tests around batching, caching, and validation.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,6 +44,18 @@ DOWNLOAD_MEDIA=true
 # Files larger than this will be skipped
 MAX_MEDIA_SIZE_MB=100
 
+# Maximum usable filename byte budget for downloaded media (the file_id prefix
+# and extension are always preserved; only the decorative part is shortened).
+# Default 143 keeps names writable on Synology/eCryptfs encrypted shares, whose
+# filename encryption overhead caps components below the usual 255 bytes.
+# Raise to 255 on plain ext4/xfs/btrfs if you prefer longer decorative names.
+# MEDIA_MAX_FILENAME_BYTES=143
+
+# Stop retrying a media file's download after this many failed attempts, so a
+# permanently-unwritable file can't tax every backup run forever. Re-requesting
+# the download (e.g. via VERIFY_MEDIA or the admin panel) resets the counter.
+# MEDIA_MAX_DOWNLOAD_ATTEMPTS=5
+
 # Timeout for media downloads (in seconds). 0 disables the timeout.
 # Default is 3600 (60 minutes).
 # DOWNLOAD_TIMEOUT_SECONDS=3600
@@ -227,6 +239,10 @@ DB_PATH=/data/backups/telegram_backup.db
 # VIEWER_USERNAME=admin
 # VIEWER_PASSWORD=your_secure_password
 # AUTH_SESSION_DAYS=30
+#
+# ALLOW_ANONYMOUS_VIEWER grants READ-ONLY access with no login: browsing and
+# search work, but nothing that writes (settings, viewer/token management,
+# deletions, etc.) is reachable without authenticating as the master account.
 # ALLOW_ANONYMOUS_VIEWER=false
 
 # --- Multi-User Access Control (v7.0.0) ---
@@ -239,7 +255,13 @@ DB_PATH=/data/backups/telegram_backup.db
 # For use behind auth proxies like Authelia, Authentik, or Keycloak.
 # The proxy authenticates the user and forwards their identity in a header.
 # Set AUTH_PROXY_HEADER to enable (e.g., "Remote-User" or "X-Forwarded-User").
-# IMPORTANT: Only enable if your reverse proxy strips this header from client requests.
+#
+# !!! WARNING !!!
+# Your reverse proxy MUST strip or overwrite this header on EVERY inbound
+# request before it reaches this app. If a client can set the header itself
+# (because the proxy passes it through untouched), anyone can impersonate any
+# user — including an admin — with a single request header. This is a full
+# authentication bypass, not a hardening nicety.
 # AUTH_PROXY_HEADER=Remote-User
 
 # Comma-separated list of usernames that get admin (master) role.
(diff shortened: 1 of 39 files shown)


GeiserX:
Thanks @coderabbitai — addressed in b9aefac:

- **Push endpoint validation timeout (Major):** fixed. The blocking `getaddrinfo` in `validate_push_endpoint` now runs under a 5s `asyncio.wait_for`, and a timeout fails fast with the existing `400`, so a hostile/slow hostname can't occupy an executor worker unbounded on this (anonymously-reachable) endpoint.
- **Realistic test password (Minor):** fixed — `test@value/here` per the repo's test-data guideline.

Declining the `except ValueError, TypeError:` finding (adapter.py) — it is **not** a Python 3 `SyntaxError` here. This project targets Python 3.14 only (`requires-python = ">=3.14"`), where [PEP 758](https://peps.python.org/pep-0758/) makes parenthesis-free `except` with multiple types valid. Concretely: `ruff format --target-version py314` **rewrites** the parenthesized form back to this comma form (verified), so `except (ValueError, TypeError):` would fail `ruff format --check` in CI; and the CI `test` job imports these modules and runs the full suite green, which a real `SyntaxError` would prevent. Leaving it as ruff's canonical form.

(The docstring-coverage warning is noted; not adding docstrings across the diff in this PR.)

PR #220, opened 2026-07-18
feat(reactions): real-time reaction capture with retain-on-removal (#219)

### LINKED ISSUE #219: [Bug]: **Title:** Reactions are only captured by the scheduled backup, never by the real-time listener
### Version

7.22.0 (backup + viewer), PostgreSQL 17, Docker Compose

### Deployment Method

Docker Compose

### Database Type

PostgreSQL

### Bug Description

Message reactions are persisted only by the scheduled backup path. The real-time
listener does not handle reactions at all, so a reaction is invisible to the
archive until the next scheduled run.

Grepping the running image confirms the asymmetry:
- grep -i reaction /app/src/listener.py        -> no matches at all
- grep -i reaction /app/src/telegram_backup.py -> extraction at ~line 1617
  (message.reactions.results), persistence at ~lines 1160-1182 via
  insert_reactions()

Impact:
1. Reaction data lags by up to the SCHEDULE interval (6h in my config).
2. A reaction added and then removed between two sweeps appears never to be
   recorded at all, with no trace it existed. (Inferred from the above; I have
   not explicitly tested add-then-remove.)

Viewer symptom, which is how I found this: a reaction updates edit_date without
changing text, so on_message_edited updates the message but writes no
message_versions row. The viewer then renders a bare, non-clickable "edited"
marker with no reaction shown. In the native Telegram app the same message shows
a heart and no edit history. It looks like a phantom edit.

### Expected Behavior

With ENABLE_LISTENER=true, reactions should be captured in real time like new
messages, edits, and deletions are, rather than waiting for the next scheduled
backup.

Suggested: a LISTEN_REACTIONS option registering a raw UpdateMessageReactions
handler, ideally with retain-on-removal semantics mirroring DELETION_MODE=soft,
so a removed reaction is preserved rather than silently dropped.

Separately, a message whose edit_date changed only because of a reaction ideally
should not surface as "edited" in the viewer, since no text version exists.

### Steps to Reproduce

1. Run with ENABLE_LISTENER=true, LISTEN_EDITS=true, SCHEDULE=0 */6 * * *,
   PostgreSQL backend.
2. Post a text message in a group that is in scope.
3. Add an emoji reaction to it from another account, shortly after posting.
4. Immediately query:
   SELECT * FROM reactions WHERE chat_id = <chat> AND message_id = <id>;
   -> 0 rows.
5. Query the message:
   SELECT id, date, edit_date FROM messages WHERE id = <id>;
   -> edit_date IS updated (so the edit event did reach the listener), but no
      reaction was stored and no message_versions row was written.
6. Open the chat in the viewer -> message shows a non-clickable "edited" marker,
   no reaction rendered.
7. Trigger a scheduled/startup sweep (I used: systemctl restart on the compose
   stack) -> the reaction row now appears and the viewer renders it.

In my case: message posted 18:18:44 local, reaction ~18:19, edit_date 18:19:08,
reactions table empty until a restart-triggered sweep.

### Logs

```shell
No errors are logged at any point -- the listener reports healthy throughout.
That is part of the problem: there is no diagnostic signal.

Evidence is in the data rather than the logs. Histogram of reactions.created_at
for one group, showing capture only on cron boundaries (0 */6 * * * = 05/11/17/23
UTC), with nothing in between apart from restart-triggered sweeps:

         hr          | count
---------------------+-------
 2026-07-18 17:00:00 |    27
 2026-07-18 11:00:00 |    30
 2026-07-18 06:00:00 |     1     <- restart-triggered sweep
 2026-07-18 05:00:00 |    40
 2026-07-17 23:00:00 |   136
 2026-07-17 19:00:00 |    10     <- restart-triggered sweep

Related viewer-side symptom in the same group -- messages with edit_date but no
preserved version, i.e. reaction-driven edit stamps:

 has_versions | count
--------------+-------
 f            |   300
 t            |    31
```

### Environment

Telegram-Archive: 7.22.0 (backup + viewer, pinned images
  drumsergio/telegram-archive:7.22.0, drumsergio/telegram-archive-viewer:7.22.0)
Database: PostgreSQL 17 (official postgres:17 image)
Deployment: Docker Compose on Ubuntu 24.04 (VM), behind Caddy reverse proxy
Chat type: group (legacy basic group)
Relevant config:
  ENABLE_LISTENER=true
  LISTEN_NEW_MESSAGES=true
  LISTEN_NEW_MESSAGES_MEDIA=true
  LISTEN_EDITS=true
  LISTEN_DELETIONS=true
  DELETION_MODE=soft
  LISTEN_CHAT_ACTIONS=true
  SYNC_DELETIONS_EDITS=false
  MASS_OPERATION_THRESHOLD=10000
  SCHEDULE=0 */6 * * *
  CHAT_TYPES=private,groups (+ GROUPS_INCLUDE_CHAT_IDS)

### Checklist

GeiserX:
## Summary

Fixes #219. Reactions were persisted **only by the scheduled backup**, so they lagged by up to the whole `SCHEDULE` interval, and a reaction added-then-removed between two sweeps left no trace. Separately, a reaction bumps a message's `edit_date` server-side, which surfaced a phantom non-clickable "edited" marker.

This adds an **opt-in real-time reaction capture** path and reworks reaction storage so both the listener and the backup converge on the same idempotent reconcile.

## What changed

**Real-time capture (`LISTEN_REACTIONS`, default off)**
- A raw `UpdateMessageReactions` handler (mirrors the existing pin handler) reconciles reactions the moment they change. Opt-in because push delivery to a user client is best-effort (no gap recovery — Telegram itself recommends polling), so the scheduled backup stays the reconciliation backstop.
- Bursts on a popular message are **debounced** per `(chat, message)` (`REACTION_DEBOUNCE_SECONDS`, default 1.5) into one write + one broadcast, and reactions are kept **out of** the shared mass-operation limiter so a reaction storm can't rate-limit edits/deletions.

**Storage: aggregate + retain-on-removal**
- Reactions are stored as **per-emoji aggregate counts** (one row per emoji). Per-user attribution is intentionally *not* persisted: a user client only sees a small, unreliable `recent_reactions` preview, so "who removed their reaction" cannot be tracked accurately (and broadcast channels are aggregate-only by Telegram policy). This also matches what the archive already stored in practice.
- A removed reaction is **retained** via a `removed_at` tombstone rather than dropped (mirroring `DELETION_MODE=soft`); the live count excludes tombstoned rows.
- The new `reconcile_reactions` replaces the old full delete-then-reinsert. It preserves `created_at` across re-scans (so "first seen" is meaningful — the old path reset it every run), reconciles removals **down to zero**, skips messages that aren't archived (no synthetic stub → no FK error on PostgreSQL), and never upserts on the nullable `user_id` unique constraint (which would grow unbounded).

**Phantom "edited" fix (ingestion, both paths)**
- A same-text edit no longer advances the stored `edit_date` — fixed in both `_should_apply_edit_text` (listener) and `_should_apply_upsert_text` (backup/gap-fill/import), so a reaction-only change never looks like an edit and the fix can't be re-introduced by the next sweep. Real text edits still record a version and show `edited(N)`; the fix is at ingestion, not the viewer, so genuine edits (including pre-versioning / formatting-only) are never hidden.

**Viewer**
- Reactions already ride the existing 3s poll (which field-merges them); a lightweight WS `reaction` event makes updates instant. No new rendering — the reactions block already renders the aggregate shape.

**Schema**
- Migration `018` adds `reactions.removed_at` and a chat-first `(chat_id, message_id)` index; idempotent, with entrypoint stamping for both PostgreSQL and SQLite.

## Tests
Full suite **2191 passed**, `ruff check` + `ruff format --check` clean.
- `test_reactions_realtime.py` (new, real SQLite): reconcile add / count-change / tombstone-on-removal / zero-clear / revive-keeps-created_at / hard-delete / legacy-multi-row collapse / skip-unarchived; the `extract_reactions` helper; a reaction-only-edit e2e that asserts `edit_date` is not bumped.
- `test_listener_reactions.py` (new): handler buffers + coalesces to the latest snapshot, disabled when the flag is off, untracked-chat + excluded-topic skips, flush reconciles + broadcasts, error isolation, PII-safe.
- `test_migration_018.py` (new): idempotent up/down, create_all-provisioned DB stays clean.
- Updated the two deliberately-pinned adapter tests (same-text now a no-op — the behavior being fixed) and the backup reaction tests (aggregate extraction + reconcile).

## Notes / scope
- Best-effort by design; the sweep reconciles every processed message (including empty snapshots) so removals-to-zero persist even with the listener off.
- Per-user "who reacted / who removed" history is explicitly out of scope (not reliably obtainable on a user client); documented rather than half-implemented.
- Reactions are still not round-tripped by export/import (pre-existing; not addressed here).

Closes #219

## Summary by CodeRabbit

* **New Features**
  * Added opt-in real-time Telegram reaction capture with configurable debouncing.
  * Reactions now appear in the viewer immediately as per-emoji aggregate counts.
  * Reaction removals are preserved and reflected accurately, including changes to zero.

* **Bug Fixes**
  * Prevented reaction-only updates from creating false “edited” indicators.
  * Improved reaction synchronization across live updates and scheduled backups.

* **Documentation**
  * Documented new reaction settings and included release notes for version 7.23.0.
  * Added migration guidance for existing databases.


GeiserX:
Thanks @coderabbitai — good catches. Addressed the substantive ones:

**Fixed**
- **Serialize reconcile per message** (`reconcile_reactions`): now locks the parent `Message` via `_load_message_for_update` before reading reactions, so the listener and the scheduled backup can't both read "no rows" and insert duplicate NULL-`user_id` aggregate rows (which `uq_reaction` wouldn't dedup). Doubles as the FK-existence guard.
- **Zero/negative counts treated as absent**: `count <= 0` entries are now excluded from `desired`, so an explicit `0` tombstones/removes the emoji instead of reviving it.
- **Retry after sequence reset**: on the PG `reactions_id_seq` collision path, we now reset the sequence and retry the reconcile once with the same snapshot (bounded, `_after_seq_reset` guard) instead of returning `noop` and dropping the authoritative state.
- **Extraction failure no longer means "empty"**: `extract_reactions` now returns `None` on unexpected shape (vs `[]` for a valid empty snapshot). Both callers (`_commit_batch`, the listener handler) skip reconciliation on `None`, so transient Telethon shape drift can't tombstone valid reactions. Added tests for both the `None`-skip and `[]`-zero-clear paths.
- **Migration-018 stamping now requires both artifacts**: the `018` stamp in both the PostgreSQL and SQLite blocks of `entrypoint.sh` now checks `reactions.removed_at` **and** the `idx_reactions_chat_message` index, so a partial schema can't skip creating the index.
- **Strengthened `test_migration_018`**: inserts a row *before* the upgrade to prove pre-existing rows are preserved with `removed_at` NULL (not just the new-column default).
- The `docs/AUTOPILOT-WORKLOG.md` / `docs/GOAL.md` files were committed by accident (local scratch) and have been removed from the branch — the markdown-lint nits on them are moot.

**Declining, with reasons**
- **"Backup does not reconcile existing messages" (bounded reaction sweep):** this is pre-existing, incremental behavior (the old reaction path only ran on newly-processed messages too), not a regression introduced here. Real-time capture is exactly what the new `LISTEN_REACTIONS` listener provides; a periodic re-sweep of *already-archived* messages to catch missed reaction events is a separate enhancement with real API-cost implications (re-fetching reactions for historical messages), so it's out of scope for this fix. Noting it as a follow-up.
- **Convert new test files to `unittest.TestCase` / add type hints to migration-test helpers:** declining for consistency — `test_reactions_realtime.py` and `test_migration_018.py` deliberately mirror their sibling files `test_jump_window_backend.py` and `test_migration_017.py`, which are pytest-function style with untyped helpers. Matching the files they're modeled on keeps the suite coherent.

(The docstring-coverage warning is noted; not adding docstrings across the diff here.)

Issue #222, opened 2026-07-19
[Bug]: LISTEN_CHAT_ACTIONS writes phantom messages rows with non-message IDs (mislabeled photo_removed) on every join/leave


625801:
### Version

7.23.0

### Deployment Method

Docker Compose

### Database Type

PostgreSQL

### Bug Description

With LISTEN_CHAT_ACTIONS=true, every group join/leave event writes two extra
rows into `messages` with an `id` that is not a Telegram message ID, labeled
action_type "photo_removed". They render in the viewer as duplicate
"<user> removed the group photo" entries directly above the correct
"joined via invite link" entry.

The IDs are time-derived, not message IDs. In one chat, real message IDs are
around 226,000, while every phantom row has an ID between 105,704,152 and
242,900,425. The value decrements exactly 1 per millisecond of wall-clock time:

  id 164862865 @ 2026-07-18 17:00:47.792241
  id 164857762 @ 2026-07-18 17:00:52.895731
    -> 5.103490 s apart, IDs differ by 5,103

  id 164862865 @ 2026-07-18 17:00:47.792241
  id 105704152 @ 2026-07-19 09:26:46.505183
    -> 59,158.713 s apart, IDs differ by 59,158,713

Exact to the millisecond across a 16-hour span, so the stored ID is
(constant - current_time_ms) rather than anything from Telegram.

The split is clean by action_type - only photo_removed is affected:

       action        | phantom_ids | normal_ids |    min    |    max
  -------------------+-------------+------------+-----------+-----------
   chat_delete_user  |           0 |          8 |    224736 |    225584
   chat_joined_by_link|          0 |         26 |    223642 |    226971
   photo_removed     |          71 |          0 | 105704152 | 242900425

26 joins + 8 deletes = 34 real chat actions, against 71 phantom rows, i.e. about
two per event. The pairing is visible in the timestamps: a genuine join at
06:23:55 produced phantoms at 06:23:55.951072 and 06:23:55.926435, ~25 ms apart.

The underlying capture is correct - the real service row has
{"service_type": "service", "action_type": "chat_joined_by_link"}. Only the extra
rows are wrong.

Impact:
- Junk rows in `messages` that inflate message counts in Backup Statistics
- Phantom "removed the group photo" entries in the rendered timeline
- Nothing upstream will ever reconcile a row whose ID does not correspond to a
  real Telegram message, so they appear permanent

### Expected Behavior

A group join/leave should produce exactly one service message row, with the real
[...]

PR #223, opened 2026-07-19
fix: archive chat actions with real IDs and capture own-account reactions (#221, #222)

### LINKED ISSUE #221: [Bug]: Listener never captures reactions made by the archive's own account
### Version

7.23.0

### Deployment Method

Docker Compose

### Database Type

SQLite (default)

### Bug Description

First, thank you for the fast turnaround on #219/#220 - reported and shipped the
same day. After a day running 7.23.0 in production: other members' reactions
land within seconds instead of waiting for a cron boundary, removals are
tombstoned, and the phantom "edited" marker is gone. The reconcile rework is a
clear improvement.

One behavior I did not expect: the real-time listener never captures reactions
made by the account the archive itself is logged in as.

Telegram does not appear to push UpdateMessageReactions back to the session that
performed the action - that client already knows - so the archive, being logged
in as me, is blind to exactly my own reactions. Other members' reactions on the
same messages in the same chat arrive live and correctly.

Self-reactions therefore fall back to pre-7.23.0 behavior: captured only if the
message happens to fall inside the scheduled sweep's incremental window.

This compounds with the bounded re-sweep noted as out of scope in #220. Since
the sweep fills its batch with min_id=last_message_id, a self-reaction on an
older message is invisible to both paths - the listener never receives it, and
the sweep never revisits the message. This is a permanent, silent gap (verified:
reacted to a 2-day-old message, ran a full sweep, no row was ever written), and
it applies specifically to the archive owner's own reactions, which are arguably
the ones a personal archive most wants to keep.

Happy to test any of the options below on my instance - two accounts on one host
(~40 and ~490 chats, PostgreSQL), so I can report API cost at both scales.

### Expected Behavior

With LISTEN_REACTIONS=true, reactions by the archive's own account should be
captured like any other reaction, or the limitation should be documented.

Options, in rough order of cost:
- Document it (main ask - only discoverable by testing)
- Opportunistically re-fetch a message shortly after the archive's own account
  is observed acting in that chat
- The bounded re-sweep from #220 (messages that already have reaction rows, or
  a configurable lookback such as REACTION_RESWEEP_DAYS, default off), which
  would cover this as a side effect

### Steps to Reproduce

1. Run 7.23.0 with ENABLE_LISTENER=true and LISTEN_REACTIONS=true.
2. Have another member react to a message -> row appears in `reactions` within
   seconds. Working as intended.
3. From the account the archive is logged in as, react to a recent message
   -> no row, nothing in listener logs. Reaction is visible in the native app.
4. Trigger a sweep -> the row appears, created_at stamped at sweep time
   (i.e. recovered by the sweep, not the listener).
5. From the same account, react to a message older than the sweep's incremental
   window (mine: id 223649, ~2 days old, ~3,400 message ids behind the newest)
   -> no row after a full sweep cycle. Neither path ever captures it.

### Logs

```shell
No errors at any point; the listener reports healthy. Nothing matching
"reaction" is logged for self-reactions, while other users' reactions are
handled normally.

Recent message, reacted by the archive's own account (steps 3-4):
- Immediately after reacting: 0 rows in `reactions`.
- After a sweep:
   id  | message_id | emoji | count |         created_at         | removed_at
  7776 |     227105 | fire  |     1 | 2026-07-19 08:51:22.051752 |
  created_at is the sweep time, not the reaction time.

Older message, reacted by the archive's own account (step 5):
  SELECT ... FROM reactions WHERE chat_id = <chat> AND message_id = 223649;
  -> 0 rows, before and after a full sweep.
```

### Environment

Telegram-Archive: 7.23.0 (backup + viewer)
Database: PostgreSQL 17
Chat type: group (legacy basic group)
Relevant config:
  ENABLE_LISTENER=true
  LISTEN_REACTIONS=true
  REACTION_DEBOUNCE_SECONDS=1.5
  LISTEN_NEW_MESSAGES=true
  LISTEN_EDITS=true
  LISTEN_DELETIONS=true
  DELETION_MODE=soft
  SYNC_DELETIONS_EDITS=false
  SCHEDULE=0 */6 * * *

### Checklist

GeiserX:
Fixes #221 and #222.

## #222 — phantom chat-action rows with fabricated IDs

Root cause (present in every release v5.3.1 → v7.23.0):

- The `on_chat_action` classifier's second branch (`photo is None and not new_photo`) was a catch-all that labeled **every** non-photo chat action (joins, leaves, title changes) as `photo_removed` — the `user_joined`/`user_left`/`user_added`/`user_kicked`/`title_changed` branches were dead code (0 rows with those labels exist across 2.95M messages in two production DBs). Genuine photo removals were misclassified as `photo_changed`.
- Rows were written with a fabricated ID: `-int(time.time() * 1000) % 2147483647` — a positive 31-bit value that decrements 1/ms and wraps every ~24.8 days, interleaving real message IDs (production forensics found fabricated IDs as low as 22M in chats whose real IDs are in the hundreds). One join produces ~2 ChatAction events (Telethon builds one from `UpdateChatParticipantAdd` and one from the `MessageService`, dispatched as parallel tasks), hence ~2 phantom rows per join.

The fix:

- Service rows are written **only** when the event carries a real `MessageService` (`event.action_message`), using its real `id`/`date`/`sender`, classified via the shared `service_action_type()` sweep vocabulary. Participant-only sync updates (no service message) write no row — matching Telegram's own UI. `events.NewMessage` skips `MessageService`, so this handler remains the only real-time source of service rows.
- Photo semantics fixed per Telethon source: `new_photo=True, photo=None` → removed; `photo=Photo` → changed.
- A new shared `service_message_text()` helper generates the human-readable text for **both** the listener and the scheduled sweep (sweep-archived service rows previously stored empty text and rendered as blank pills).
- **Migration 019** deletes all historical phantom rows: parse-free `LIKE` matching of the 7 legacy `action_type` names on the TEXT `raw_data` column (both JSON serialization eras, escaped underscores, no id/service_type conditions — forensics found real phantoms with low IDs and without the `service_type` key). Dependent `reactions`/`media`/`message_versions` rows are deleted explicitly first (the reactions FK has no cascade). Invalidates `cached_stats` + `stats_calculated_at` so the viewer recomputes correct counts immediately on next start. Data-only migration: intentionally absent from the entrypoint stamp ladders (documented there) so `alembic upgrade head` always runs it; naturally idempotent. Validated against a full copy of a production DB: deletes exactly the 422 forensically-confirmed phantoms out of 1.78M rows.

## #221 — reactions by the archive's own account never captured live

Telegram does not reliably push `UpdateMessageReactions` to the archive's session for reactions made by the same account from another device. Mitigations, all preserving the #220 aggregate-only model:

- **Edit harvest**: Telegram delivers some reaction changes as genuine edit events (Telethon #4635, maintainer-confirmed). The edit handler now feeds reaction snapshots carried on edit payloads into the shared reaction debounce buffer — before the text-noop early return, so reaction-only edits (unchanged text) still reconcile without bumping `edit_date` (the #219 phantom-edit guard is untouched). `min` payloads are skipped (documented to omit the current user's own reactions — reconciling one could falsely tombstone a just-added self-reaction). New messages arriving with reactions attached are harvested the same way.
- **Bounded re-sweep (opt-in)**: `REACTION_RESWEEP_DAYS=N` re-checks the last N days of messages per chat on every scheduled sweep via batched `messages.getMessagesReactions` (100 ids/request, `get_messages` fallback for rejected chunks), capped by `REACTION_RESWEEP_MAX_PER_CHAT` (default 500 ≈ 5 requests/chat/sweep). Default `0` (disabled). Only ids echoed back are reconciled — absence is never treated as "zero reactions".
- **Piggyback**: `SYNC_DELETIONS_EDITS=true` already fetches full messages and discarded their reactions; they are now reconciled at zero extra API cost.
- **Docs**: README section "Reactions made by your own account" documents the limitation and both knobs.

## Testing

- 2241 tests pass (36 new for #221 paths, 22 new for #222 classifier/helper, 9 for migration 019); `ruff check` + `ruff format --check` clean.
- Migration validated against a full production-DB copy (1,777,032 rows): predicate matches exactly the 422 known phantoms; sweep-vocabulary service rows and all normal rows untouched.
- Live deployment validation on two production instances planned before release.

## Summary by CodeRabbit

* **New Features**
  * Added optional reaction “resweep” to recover self-reaction counts with configurable lookback days and per-chat caps.
  * Improved reaction capture by buffering edit/new-message reaction vectors and reconciling reactions during sync.
  * Enhanced service-message text rendering (joins, member changes, title edits, photo updates).

* **Bug Fixes**
  * Improved chat-action archiving using the real Telegram service message metadata.
  * Better preservation/synthesis of service text during historical backups and reaction reconciliation.

* **Documentation**
  * Documented new environment variables and how/when to use resweep settings.

* **Tests**
  * Added coverage for reaction resweep, chat-action handling, and migration 019 behavior.

* **Chores**
  * Improved DB engine logging safety (hide bound parameters) and clarified entrypoint migration stamping.


GeiserX:
Ran an 8-perspective review panel (logic, architecture, security/PII, simplicity, tests, concurrency, data-integrity, error-handling) over the full diff, with adversarial verification of every critical/high finding. Results, addressed in d06ddc0:

**Fixed**
- **Wrong subject in synthesized service texts (high, confirmed)**: the sweep named the acting admin ("Mod was removed…") instead of the affected user for `ChatAddUser`/`ChatDeleteUser`, and the upsert text gates would have frozen that wrong attribution permanently — including retroactively backfilling every historical blank service row. The subject is now resolved from the local users table (API fallback), degrading to "Someone …" when unknown, never to the wrong name. Self-adds read "joined the group". One shipped test had encoded the bug as expected behavior; rewritten plus 5 new cases.
- **Reaction-buffer stale-snapshot race (confirmed, reproduced)**: `on_new_message`'s awaits opened a window where its event-time snapshot could clobber a fresher one buffered by the live reaction handler. Harvesting now goes through a single helper: a missing reactions object is treated as no-information (never an authoritative empty snapshot), and the new-message producer uses `setdefault` so a fresher live snapshot always wins. Regression tests simulate the interleaving.
- PII hardening: SQLAlchemy engines now set `hide_parameters=True` so DB error logs never embed bound values (message text, chat ids); silent actor-lookup failures now log the exception class.
- Stale comments claiming the listener/sweep action vocabularies "differ by design" (they were unified by this PR) corrected; migration comment about stats caching corrected (per-chat stats sit behind a self-expiring 60s web cache).
- Test gaps closed: rate-limited edits still harvest reactions; joins don't trigger the chat-metadata refresh; real-wired `reply_to_top_id`/`reply_to_msg_id`/`is_outgoing` asserted; resweep per-request id slices asserted; the days=0 test now tests what its name says.

**Declined (with evidence)**
- *Combine the migration's 4 LIKE scans into one temp-table pass*: measured on a full copy of a 1.78M-row production DB — the entire migration ran in under ~2s, one time. Not worth cross-engine temp-table complexity.
- *Resweep MSG_ID_INVALID "retries through the full flood budget"*: factually wrong — `call_with_flood_retry` retries only `FloodWaitError`; any other RPC error re-raises immediately and triggers the fallback at once.
- *Rolling-upgrade window (old listener writing phantoms after 019 ran)*: compose deployments stop the old container before starting the new one, so the window doesn't exist there; anyone who somehow hits it can re-run the idempotent cleanup (`alembic downgrade 018 && alembic upgrade head`). A regression test pins that the fixed listener can never emit the legacy names.
- *Freshness/versioning for reconcile_reactions across its 3 callers*: reconciles are full-snapshot and idempotent; a lost race is corrected by the next observation. Versioning was deliberately rejected in the #219/#220 design as unnecessary for an aggregate model.
- *Pre-filtering resweep reconciles against a batched read*: real but bounded (opt-in feature, ≤500 row-locked no-op reconciles per chat per sweep). Noted as a follow-up optimization if anyone runs it at large scale.

Full suite: 2253 passed; ruff clean. Migration re-validated unchanged against the production copy (predicate untouched by this round).

GeiserX:
CodeRabbit round addressed:

- **.env.example wording** — fixed in the latest commit: "self-reactions are not reliably pushed" (aligned with the README and the implementation's treatment of delivery as unreliable rather than impossible).
- **Uncommenting the resweep env lines in docker-compose.yml** — declined: every optional listener knob in this compose file is a commented `${VAR:-default}` documentation line by convention (`LISTEN_REACTIONS`, `REACTION_DEBOUNCE_SECONDS`, the `MASS_OPERATION_*` block). Users uncomment what they enable, as the README documents. Uncommenting only the resweep pair would break that convention while its master switch stays commented.
- **unittest.TestCase conversion for the new test classes** — declined for the same reason as in #220: the reaction/listener test files deliberately use pytest-style classes matching their direct siblings (`test_listener_reactions.py`, `test_reactions_realtime.py`); converting one class would make the file internally inconsistent.

PR #225, opened 2026-07-20
fix: pace the reaction resweep globally and defer on FloodWait (#224)

### LINKED ISSUE #224: [Bug]: REACTION_RESWEEP trips FloodWait on getMessagesReactions (burst rate, not volume)
### Version

7.24.0

### Deployment Method

Docker Compose

### Database Type

PostgreSQL

### Bug Description

Thanks for the fast turnaround on both #221 and #222 - the phantom chat-action
cleanup in migration 019 removed 99 rows on my instance and joins now render
correctly, and the resweep works and found real discrepancies. Reporting a
pacing issue with the latter.

The resweep trips FloodWait after only a few getMessagesReactions requests,
accumulated ACROSS chats rather than within one chat, so a per-chat cap cannot
avoid it.

Two distinct FloodWait signatures appear in these logs and should not be
confused:
  A) "before resuming (last_msg_id=...)"        - ordinary message iteration
  B) "before retrying <TelegramClient object>"  - the resweep

CONTROL, REACTION_RESWEEP_DAYS=0, 429 dialogs, completed in 264s:
  Only type A, 7 occurrences, at dialogs 58/118/178/238/298/358/418 - a regular
  ~60-dialog cadence, clearly the baseline iteration limit for an account this
  size. Zero type B.

WITH RESWEEP, DAYS=3, MAX_PER_CHAT=300, fresh start after hours idle:
  [1/429] -> checked 145 ids (2 requests), reconciled 8   - OK
  [2/429] -> checked  58 ids (1 request),  reconciled 3   - OK
  [3/429] -> checked   6 ids (1 request),  reconciled 0   - OK
  [4/429] type B FloodWait 60.45s, then checked 300, reconciled 5
  [5-8]   -> 1, 6, 9, 2 ids                               - OK
  [9/429] type B FloodWait 61.02s, then checked 300, reconciled 0

So ~4 batched requests are enough to trip it, and the budget appears to refill
over roughly a minute.

Impact scales sharply with MAX_PER_CHAT. At 300 the run completes normally
(387s vs 264s with the resweep disabled) - the ~120s delta is almost entirely
the two type B sleeps, so the cost is wasted wall-clock rather than failure.
At 2000 the first dialog exhausted all five retries (89s, 24s, 61s, 59s, 152s)
and the run never advanced past dialog 1.

Note the resweep fired on only ~14 of 429 dialogs at DAYS=3, since the rest had
no messages in the window - so total request volume is low. The trip appears to
be burst rate over a short window, not aggregate volume, which is why a per-chat
cap cannot express it.

Concretely: a short sleep between getMessagesReactions batches (configurable,
defaulting to something small like 0.5s) would cost a few seconds per sweep
while removing ~120s currently spent in FloodWait backoff - i.e. faster overall,
not slower.

### Expected Behavior

The resweep should pace itself so a scheduled sweep does not trip FloodWait.
Options, in rough order of preference:
- A small configurable delay between getMessagesReactions batches
- A global per-sweep request budget rather than a per-chat cap
- Skipping the resweep for the remainder of a run after the first type B
  FloodWait (cheap safety net, complements the delay rather than replacing it)

### Steps to Reproduce

1. 7.24.0, ENABLE_LISTENER=true, LISTEN_REACTIONS=true, an account with several
   hundred dialogs and a few high-volume chats.
2. Set REACTION_RESWEEP_DAYS=3, REACTION_RESWEEP_MAX_PER_CHAT=300.
3. Start after an idle period and watch the scheduled sweep.
4. Type B FloodWaits appear on the chats whose resweep id count is large, while
   chats with a handful of ids pass instantly.
5. Set REACTION_RESWEEP_DAYS=0 and repeat - only type A FloodWaits remain.

### Logs

```shell
Type B (resweep), DAYS=3 / MAX_PER_CHAT=300:
  02:54:37 FloodWait: sleeping 60.45s ... before retrying <TelegramClient ...> (retry=1/5)
  02:55:39 FloodWait: sleeping 61.02s ... before retrying <TelegramClient ...> (retry=1/5)

Type B at MAX_PER_CHAT=2000 - all five retries on dialog 1, run never advanced:
  wait=89s (1/5), 24s (2/5), 61s (3/5), 59s (4/5), 152s (5/5)

Type A (baseline iteration, present with the resweep disabled too):
  FloodWait: sleeping 24.82s ... before resuming (last_msg_id=209691, retry=1/5)
```

### Environment

Telegram-Archive: 7.24.0 (backup + viewer)
Database: PostgreSQL 17
Account size: 429 dialogs after filtering (537 total, 88 archived),
  ~183k messages, 500 chats
Config:
  REACTION_RESWEEP_DAYS=3
  REACTION_RESWEEP_MAX_PER_CHAT=300
  ENABLE_LISTENER=true, LISTEN_REACTIONS=true
  REACTION_DEBOUNCE_SECONDS=1.5
  SYNC_DELETIONS_EDITS=false
  SCHEDULE=0 */6 * * *
  CHAT_TYPES=private,groups

### Checklist

GeiserX:
Fixes #224.

## Root cause

`messages.getMessagesReactions` is flood-limited by **burst rate per (account, method)** — not per chat — with an observed bucket of ~4-5 requests refilling over roughly a minute. Both Telethon and gotd/td key their internal flood caches purely by RPC constructor ID (no peer dimension), architecturally confirming the reporter's cross-chat-accumulation diagnosis, and the official method docs recommend ≥15s spacing. `REACTION_RESWEEP_MAX_PER_CHAT` therefore could not express the real constraint, and two further defects amplified the damage:

- The retry wrapper's 5 retries burned up to ~385s inside a live flood window on a single dialog (retrying into an active window compounds the penalty).
- A retry-exhausted `FloodWaitError` was swallowed by the generic fallback handler and fell through to `get_messages` — a *second* rate bucket with five more retries.

## The fix

- **Global pacing**: resweep API requests are spaced by `REACTION_RESWEEP_BATCH_DELAY_SECONDS` (default `2`), measured **across chat boundaries** via one per-run timestamp (the observed trip pattern is ~1 request per chat, so per-chunk spacing inside a chat would never fire). Documented honestly as burst smoothing — a delay sized to make floods impossible (~15s) would cost more than it saves.
- **Defer, don't fight**: the raw request runs without a retry wrapper; on the **first** `FloodWaitError` of a run (raw or fallback path) the remaining dialogs' resweep is deferred to the next scheduled sweep — no retry burn, no falling back onto the second bucket. The pathological 385s/dialog behavior is structurally impossible now.
- **Chat-keyed cycle cursor**: completed chats are persisted (`reaction_resweep_cycle_done` metadata key) and skipped until the cycle completes, so a deferred run's next sweep resumes with the chats that were skipped. Without this, the recency-sorted dialog order would re-sweep the same head chats every run and permanently starve the tail. A clean run clears the cursor.
- The `get_messages` fallback is now reserved for genuine non-flood raw errors (its original intent), paced identically, and limited to one flood retry.
- The confusing `before retrying <TelegramClient object>` log line is gone (that call site no longer goes through the retry wrapper).

## Behavior for the reporter's two scenarios

- `MAX_PER_CHAT=300` (normal run): requests spaced 2s apart; if the bucket still trips, one warning, remainder deferred, resumed next sweep — no ~120s of flood sleeps.
- `MAX_PER_CHAT=2000` (pathological): the first flood defers the run's resweep immediately instead of 5 retries × 2 buckets on dialog 1; the cursor spreads the window across consecutive sweeps.

## Testing

2262 tests pass (8 new: global cross-chat pacing, delay-disabled, raw-flood deferral without fallback, fallback-flood deferral, cycle marking/skip, cursor load/persist/clear, corrupt-cursor tolerance, disabled-mode no-ops); ruff clean. Live validation on a production instance planned before release.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reaction re-sweeps with configurable global request pacing to reduce FloodWait errors.
  - FloodWait-deferrred re-sweeps now persist progress and resume remaining work on the next scheduled run.
  - Prevented already completed chats from being processed again within the same cycle, including after mid-chat interruptions.
- **Documentation**
  - Documented the optional `REACTION_RESWEEP_BATCH_DELAY_SECONDS` setting, including behavior and `0` (disabled pacing).
- **Chores**
  - Updated the application version to **7.25.0**.


GeiserX:
Ran a 5-perspective review panel (logic, concurrency/state, simplicity, tests, behavior-docs) with adversarial verification. Addressed in 98bb955:

**Fixed**
- **Per-chat resume progress (medium, the load-bearing one)**: a chat whose window needed more chunks than the flood bucket flooded at the same chunk every run, was never marked done, and — via the defer gate — starved every chat behind it. The cycle cursor now records mid-chat progress (`partial` offsets) so the next run resumes exactly where the flood hit; forward progress is guaranteed for any chat size.
- **Symmetric flood deferral**: the `get_messages` fallback used `max_retries=1`, which slept a full FloodWait and retried into the live window before deferring — contradicting the raw path's semantics. It now defers immediately, no wrapper, no sleep.
- **Cursor hardening**: the persisted cursor is stamped with its window setting and timestamp and discarded on mismatch or staleness (>48h), so disabling and later re-enabling the resweep can't resurrect weeks-old coverage; finalize now runs directly after the dialog loops so a later topics/folders/statistics failure can't drop a deferral.
- Gate order (done-before-flood) fixes the deferred-count over-count; monotonic sentinel wart removed; the cursor load key is now pinned by a test (a load/persist key mismatch was invisible to the suite).
- **Version corrected to 7.25.0** — a new user-facing env var plus changed default behavior is a minor bump by this repo's semver convention, not a patch.

**Declined with reasoning**
- *Cross-process cursor races (manual backup overlapping a scheduled run)*: the cursor is best-effort catch-up state over an idempotent reconcile — a last-writer-wins clobber costs at most one cycle's efficiency, never correctness, matching the rest of `backup_all` which is likewise unsynchronized across processes.
- *Crash mid-dialog-loop discards this run's cursor progress*: inherent to persist-at-end; the previous run's cursor remains valid and the cycle self-heals on the next clean run.

2267 tests green (13 new for #224 paths), ruff clean.

GeiserX:
@coderabbitai full review

625801:
Thanks - the deferral works exactly as described: one clean warning, no sleeping,
no retries, and the `retrying <TelegramClient object>` signature is gone.

Taking you up on the offer to reopen, though, because at my scale the deferral
costs more coverage than the floods did.

My bucket is much smaller than yours - I trip after ~4 requests where your
instance absorbed 27. So a run now covers 2 chats and defers 498:

  Reaction resweep: checked 300 ids, reconciled 0
  Reaction resweep: checked 59 ids, reconciled 0
  Reaction resweep hit a getMessagesReactions FloodWait (53s); deferring the
    rest of this run's resweep
  Reaction resweep deferred for 498 dialogs after a FloodWait; 2 dialogs are
    done this cycle and the rest resume next run

Under 7.24 the same config covered my entire eligible set (~14 chats with
messages in the DAYS=3 window) in a single run - it slept ~120s across two
floods and then continued. That wall-clock cost was harmless for me; nothing in
my setup depends on sweep duration. Trading it for multi-run cycle latency is a
net loss at this scale.

A possible middle ground: the run itself lasts ~390s and the reported FloodWait
was 53s, so the bucket refills long before the run ends - message iteration
continues normally throughout. Could the resweep record the flood expiry and
resume once wall-clock passes it, still within the same run? That keeps
everything the current design protects - no blocking sleep, no retries, no
fallback onto the second bucket - while restoring per-run coverage for accounts
whose bucket is small. Deferral to the next run would then only happen if the
deadline falls after the run ends.

Happy to test a build either way. Currently on DAYS=3 / MAX_PER_CHAT=300; I'll
also try MAX_PER_CHAT=100 meanwhile, since spending fewer requests per chat
should let each run reach further down the list.

625801:
Tried MAX_PER_CHAT=100 as mentioned - it resolved the coverage problem on my
account. Two consecutive runs, the second starting ~30 min after the first:

  12 chats resweeped, no FloodWait, no deferral
  Duration 264.69s, versus a 264s baseline with the resweep disabled

The run stopped because it ran out of eligible chats, not because it flooded -
so a full cycle completes every run, at effectively zero cost. Most of my
eligible chats are small (1-9 ids), so they cost ~1 request each and a lower cap
lets the run reach all of them.

For completeness, MAX_PER_CHAT=500: 1 chat done, 499 deferred, 260s.

Two caveats before you draw conclusions from this:

- The three configs weren't measured independently. Testing order was
  300 -> 100 -> 500 with many service restarts in between, and my flood bucket
  was clearly degrading through the evening - by the end I was tripping after
  ~1 request where I'd managed 12 earlier. So 500 appearing worst may partly be
  measurement order rather than the cap value.

- I can't explain the trip condition. 12 spaced requests ran clean at 100 while
  ~5 flooded at 500 - not a simple request count, and total ids don't fit either
  (~392 clean vs ~460 flooding).

Re-measuring on a rested account with scheduled sweeps only, no restarts, and
I'll follow up once I have something trustworthy. The middle-ground suggestion
in my previous comment still stands

GeiserX:
Heard, and agreed — your middle ground is the right design. The deferral was built to never fight the limiter, but at a ~4-request bucket it over-corrects: trading ~120s of in-run wall-clock for multi-run cycle latency is a net coverage loss, exactly as your numbers show.

Shipping your proposal: on a resweep FloodWait the run records the expiry deadline and keeps going (message iteration continues as before, nothing sleeps); chats reached while the window is still cooling skip to the next-run cursor as today, and once wall-clock passes the expiry the resweep resumes **within the same run**. Full deferral only happens if the deadline outlives the run — or after a few floods in one run (safety cap for degrading buckets like the one you saw by evening). Also adding your `MAX_PER_CHAT` finding to the README as sizing guidance: smaller caps spend fewer requests per chat, letting one bucket reach further down the list.

Reopening #224 to track it; fix incoming as v7.25.1.

PR #226, opened 2026-07-21
fix: resume the reaction resweep within the run once a FloodWait expires (#224)

GeiserX:
Reopens and completes #224 per the reporter's follow-up on #225.

## Problem

The v7.25.0 defer-on-first-FloodWait protected against retry-burn but over-corrected on accounts with small flood buckets (~4 requests vs ~27 on our test instance): one flood deferred the entire remaining run, so a run covered 2 chats and deferred 498 — while the run itself continued for ~6 more minutes with the bucket long refilled (the reported wait was 53s). Net coverage loss versus the old sleep-through-floods behavior.

## Fix (the reporter's proposed middle ground, implemented faithfully)

- On a resweep FloodWait, record the expiry deadline (server-requested seconds + 2s margin) and **pause** — nothing sleeps, nothing retries, no fallback onto the get_messages bucket. Message iteration and the rest of the backup continue untouched.
- Chats reached while the window is still cooling skip to the next-run cycle cursor (as before).
- Once wall-clock passes the deadline, the resweep **resumes within the same run**. Full deferral now happens only when the window outlives the run — or after **3 floods in one run** (a repeatedly-flooding bucket is degrading and should be left alone until the next scheduled run).
- The finalize/persist condition is now exact: the cursor persists iff at least one chat was actually deferred.
- README gains sizing guidance from the reporter's measurements: on small-bucket accounts a lower `REACTION_RESWEEP_MAX_PER_CHAT` (e.g. `100`) spends ~1 request per chat and lets a single run reach every eligible chat.

No new environment variables; the resume margin and flood cap are internal constants.

## Testing

2269 tests pass (new: in-run resume after cooldown with a controlled clock, hard-defer after repeated floods, cooldown-skip counting; updated: pause semantics on both request paths, finalize keyed on actual deferral). Live validation on a production instance planned before release.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reaction re-sweeps under Telegram rate limiting, including the fallback reaction fetch path.
  - FloodWaits now pause and resume within the same run after the server cooldown (with a small margin), preserving mid-chat progress.
  - Repeated FloodWaits stop further work for the rest of the run (hard-deferred mode).
- **Documentation**
  - README and example env notes now more precisely describe FloodWait pause/resume behavior and resweep pacing, including per-chat sizing tradeoffs.
- **Tests**
  - Updated reaction resweep tests to validate cooldown-based pause/resume and deferred vs. hard-deferred behavior.
- **Chores**
  - Bumped version to **7.25.1**.


GeiserX:
CodeRabbit's wording catch fixed in the latest commit — the cursor resumes *approximately* where the previous run stopped (the window shifts between runs; the idempotent reconcile makes overlap harmless), matching what the code comments already said.

GeiserX:
Ran a 3-perspective panel (state-machine logic, adversarial behavior, tests) with adversarial verification: **zero serious findings** — the pause/resume state machine verified sound on gate order, cap-trip progress parking, exact finalize keying, double-count immunity, and clock discipline. Five doc/test findings all addressed in the latest commit: stale defer-on-first-flood narratives in .env.example, one code comment, and the test-class docstring; plus two new tests (PII guard over the flood/pause/resume log lines, and a double-flood-on-one-chat case proving the parked offset advances rather than resetting). 2271 tests green.

GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

GeiserX:
Last stale narrative fixed — the env-table row for `REACTION_RESWEEP_BATCH_DELAY_SECONDS` now matches the re-sweep section and `.env.example` (pause and resume within the run; defer only when the wait outlives the run or floods repeat). CodeRabbit has substantively reviewed this PR clean twice; merging once the CI suite is green on this docs-only commit rather than cycling its rate-limited status a third time.

Issue #229, opened 2026-07-23
[Feature]: Show sender avatars on group messages, plus a sender-info popup


625801:
### Feature Category

Web Viewer

### Problem Statement

In group chats the viewer identifies the sender with a colored name on the
first message of a run only. Consecutive messages from the same person carry no
attribution at all, and there is no avatar anywhere in the message list.

This makes senders hard to distinguish in practice. In one of my groups two
participants share a first name - different accounts, different photos - and the
display name alone can't tell them apart. Scanning a busy group also means
tracking back up to the last colored name to work out who is speaking.

The native client solves this with a round avatar at the bottom-left of each
sender's message run, which makes the speaker identifiable at a glance without
reading anything.

### Proposed Solution

In group chats the viewer identifies the sender with a colored name on the
first message of a run only. Consecutive messages from the same person carry no
attribution at all, and there is no avatar anywhere in the message list.

This makes senders hard to distinguish in practice. In one of my groups two
participants share a first name - different accounts, different photos - and the
display name alone can't tell them apart. Scanning a busy group also means
tracking back up to the last colored name to work out who is speaking.

The native client solves this with a round avatar at the bottom-left of each
sender's message run, which makes the speaker identifiable at a glance without
reading anything.

So this is a rendering change: resolve messages.sender_id to the matching file
under media/avatars/users/ and show it.

When no avatar file exists (users who never set a photo, deleted accounts), fall
back to a circle with the user's initials, as the native client does: white text
on a colored background, with the color derived deterministically from the user
ID so the same person always gets the same color. A gradient fill matches the
native look.

Worth constraining the palette so white text stays legible - i.e. exclude pale
and near-white fills, or otherwise guarantee contrast. Telegram uses a small
fixed set of saturated colors for exactly this reason. The existing colored-name
palette in the viewer may already be a usable source, provided the same contrast
constraint applies.

### Alternatives Considered

- Repeating the colored sender name on every message rather than only the first
  of a run. This is the obvious cheap fix but a poor one - it adds clutter and
  still doesn't disambiguate participants who share a display name, which is the
  actual problem.
- Showing the sender's @username alongside the name. Helps disambiguate, but
  many users have no username, and it's still text to read rather than
  recognize.
- Avatars in the chat list only. Useful separately (media/avatars/chats/ exists
  for this), but doesn't address identifying speakers within a group.

None of these give the at-a-glance recognition an avatar does, and all of them
leave the already-downloaded avatar files unused.

[...]

PR #230, opened 2026-07-23
feat: group-migration handling (#228) + sender avatars (#229)

### LINKED ISSUE #228: [Bug]: Basic group -> supergroup migration silently stops capture (chat_migrate_to recorded but not acted on)
### Version

7.25.1

### Deployment Method

Docker Compose

### Database Type

PostgreSQL

### Bug Description

First, thanks for 7.25.1 - the same-run resume works exactly as described:

  10:49:31 Reaction resweep hit a getMessagesReactions FloodWait (197s);
           pausing, will resume once it expires
  10:53:01 Reaction resweep cooldown elapsed; resuming within this run

No sleeping, no retries, message iteration untouched. My account's flood window
has escalated to ~197s from repeated testing yesterday, so that particular run
still ended with 0 dialogs done - the cooldown outlived the useful part of the
run. That's my bucket, not the fix. I've parked the resweep for ~24h to let the
penalty decay and will report properly measured numbers from a rested account.

Separate issue, unrelated to reactions:

When a tracked basic group migrates to a supergroup, the archive records the
chat_migrate_to service message and then silently stops capturing that chat. The
new -100 id isn't in GROUPS_INCLUDE_CHAT_IDS, so nothing is captured, and nothing
warns - no error, nothing above INFO. I only noticed ~14 hours later because the
chat had visibly stopped updating in the viewer.

Adding the new id manually backfilled cleanly - old chat ends 06:50:03 UTC, new
chat begins 06:50:02, no gap - so nothing was permanently lost in my case, but
only because I happened to check.

Note the split itself is unavoidable: the new supergroup returns only
post-migration history via the API. Telegram's own web clients differ here -
web.telegram.org/k renders the two as one continuous chat, while
web.telegram.org/a shows only post-migration messages with a "Migrated from
<title>" system message. So stitching is a client-side choice, not something the
API provides.

Escalating options:
- Log a WARNING when chat_migrate_to is seen for a chat in scope
- Auto-follow the pointer: add the migrated-to id to the effective scope
- Link the two chats in the viewer so they render as one thread. Both directions
  are already stored - chat_migrate_to on the old chat, channel_migrate_from on
  the new one (which also carries the title in new_title) - so the data is there.

The first alone would have prevented the silent stop.

### Expected Behavior

A tracked chat that migrates should not silently stop being captured. At minimum
a WARNING naming the old and new chat ids, so the operator knows to update
GROUPS_INCLUDE_CHAT_IDS. Ideally the migration pointer is followed automatically.

### Steps to Reproduce

1. Track a legacy basic group via CHAT_TYPES=private,groups and
   GROUPS_INCLUDE_CHAT_IDS=<old id>.
2. Trigger a migration to supergroup (in my case an admin enabled a
   join-by-request invite link, which is supergroup-only).
3. Telegram creates a new -100 id and writes chat_migrate_to on the old chat.
4. Capture for that chat stops. Backups continue normally for every other chat,
   no errors, no warnings.
5. The old chat's last message and the migration marker are both stored:
     {"service_type": "service", "action_type": "chat_migrate_to"}
   and the new chat's first message is:
     {"service_type": "service", "action_type": "channel_migrate_from",
      "new_title": "..."}
   but nothing links them and nothing surfaces the change.

### Logs

```shell
Nothing is logged for the migration itself. The only trace is in the data.

Old chat, final rows:
  227972 | 2026-07-21 06:50:03 | {"service_type": "service",
                                  "action_type": "chat_migrate_to"}
  227971 | 2026-07-21 06:38:03 | Николай joined the group

New chat, first row (id restarts at 1):
  1 | 2026-07-21 06:50:02 | {"service_type": "service",
                             "action_type": "channel_migrate_from",
                             "new_title": "..."}

After manually adding the new id:
      chat_id     | count |         min         |         max
  ----------------+-------+---------------------+---------------------
   -100XXXXXXXXXX |   152 | 2026-07-21 06:50:02 | 2026-07-21 16:42:44
     -XXXXXXXXXX  |  4187 | 2026-07-17 18:41:02 | 2026-07-21 06:50:03
```

### Environment

Telegram-Archive: 7.25.1 (backup + viewer)
Database: PostgreSQL 17
Affected chat: legacy basic group migrated to supergroup; ids redacted
Relevant config:
  ENABLE_LISTENER=true
  LISTEN_CHAT_ACTIONS=true
  LISTEN_REACTIONS=true
  REACTION_RESWEEP_DAYS=0 (temporarily parked)
  REACTION_RESWEEP_MAX_PER_CHAT=100
  REACTION_RESWEEP_BATCH_DELAY_SECONDS=2
  CHAT_TYPES=private,groups (+ GROUPS_INCLUDE_CHAT_IDS)
  SCHEDULE=0 */6 * * *

### Checklist

GeiserX:
Resolves #228 and #229 (both reported by @625801).

## #228 — group→supergroup migration silently stopped capture

When a tracked **basic group** migrated to a **supergroup**, the new `-100…` id was outside `GROUPS_INCLUDE_CHAT_IDS` and capture stopped with nothing above INFO. Root cause confirmed against the pinned Telethon 1.43.2 source: `events.ChatAction.build()` has no branch for `MessageActionChatMigrateTo`/`ChannelMigrateFrom`, and `events.NewMessage` skips service messages — so the migration is invisible to both live handlers, and only the scheduled sweep can ever see it.

- **Persist the pointer** in `raw_data` (`migrate_to_id` / `migrate_from_id`). `MessageActionChatMigrateTo` carries `.channel_id` (never a title), so the id was previously dropped and the archived marker alone couldn't yield the new chat.
- **Always-on WARNING** (the reporter's minimum, option a): `_reconcile_migrations` at the end of `backup_all` detects migrated in-scope groups via `entity.migrated_to` (primary) plus a stored-marker fallback for migrations that happened while offline, and emits a **counts-only** warning — deduped per run, permanently suppressed once the new id is followed or excluded. This alone ends the silent stop.
- **Opt-in auto-follow** (option b): `FOLLOW_CHAT_MIGRATIONS` (default **off**) persists followed ids in the metadata KV (no schema change) and injects them into effective scope for **both** the sweep dialog filter and the listener's tracked set, backfilling the new supergroup in the same run. Default-off preserves the include-list allow-list contract; the warning gives every operator the info to opt in deliberately.
- **Viewer banner** (option c): a migrated chat links to its counterpart using the persisted pointer.

Note: the history split is unavoidable (the API only returns post-migration history for the new supergroup) — this matches the reporter's own gapless observation.

## #229 — sender avatars + initials fallback in the group viewer

Group messages showed a colored sender name only at the start of a run — no avatar, and two participants sharing a first name were indistinguishable.

- **Initials circle at each sender run** (group chats only): a deterministic monogram on a darker gradient (`hsl(h,65%,27%→18%)`) derived from the same hue as the sender name, **verified ≥4.5:1 white-on-fill across all 360 hues** (the bright name palette fails that, which is why the fill is separate). Nameless/deleted senders render `?`, not `DA`.
- **Real member photos where they already exist**: the media ACL is fixed so a member's user avatar (`avatars/users/{user_id}`) is served to a viewer authorized for a chat that user spoke in — the old chat-id-keyed check 403'd every member avatar. The `` falls back to the initials circle on absent/error.
- **Deferred honestly**: proactive downloading of member avatars that aren't already on disk (~half, coincidentally present from being 1:1 contacts) is a follow-up — it needs a flood-gated pipeline and is out of scope here. Only pre-existing files are served.
- The sender-info popup (the issue's secondary ask) is a noted follow-up; the initials + color already solve the stated disambiguation problem.

## Testing

2320 tests pass (52 new across `test_chat_migration.py` and `test_sender_avatars.py`); ruff check + format clean. #228 is validated by unit tests + synthetic fixtures (a live migration can't be forced); #229 render is verified live.

## Summary by CodeRabbit

* **New Features**
  * Detects and handles group → supergroup migrations, with optional auto-follow via `FOLLOW_CHAT_MIGRATIONS=true`.
  * Persists adopted migrated supergroup IDs to automatically widen backup/live capture scope.
  * Adds a migration banner and improves sender avatar/initials rendering in the chat UI.
* **Bug Fixes**
  * Tightens media/avatar ACL checks with stronger path containment and “fail closed” visibility probing.
* **Documentation**
  * Updates README and environment examples with migration behavior and configuration details.
* **Tests**
  * Adds comprehensive migration and sender-avatar/ACL test coverage.
* **Chores**
  * Bumps version to 7.26.0.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -174,6 +174,13 @@ PRIVATE_INCLUDE_CHAT_IDS=
 GROUPS_INCLUDE_CHAT_IDS=
 CHANNELS_INCLUDE_CHAT_IDS=
 
+# Group → supergroup migrations (#228): when a tracked basic group is upgraded to
+# a supergroup it gets a brand-new id and capture would otherwise silently stop.
+# The scheduled backup always warns (count-only). Set FOLLOW_CHAT_MIGRATIONS=true
+# to instead adopt the new supergroup id automatically (persisted and merged into
+# the backup + listener scope). Default false — warning only.
+# FOLLOW_CHAT_MIGRATIONS=false
+
 # ==============================================================================
 # REAL-TIME LISTENER
 # ==============================================================================
--- a/README.md
+++ b/README.md
@@ -284,6 +284,7 @@ The **Scope** column shows whether each variable applies to the backup scheduler
 | `GROUPS_INCLUDE_CHAT_IDS` | - | B | Force-include specific groups |
 | `CHANNELS_EXCLUDE_CHAT_IDS` | - | B | Exclude specific channels |
 | `CHANNELS_INCLUDE_CHAT_IDS` | - | B | Force-include specific channels |
+| `FOLLOW_CHAT_MIGRATIONS` | `false` | B | Automatically adopt the new supergroup id when a tracked basic group is upgraded to a supergroup, so capture continues without editing include lists. When off, the sweep only warns. See [Group → supergroup migrations](#group--supergroup-migrations) |
 | **Real-time Listener** | | | See [Real-time Listener](#real-time-listener) below |
 | `ENABLE_LISTENER` | `false` | B | **Master switch** — enables all `LISTEN_*` features below |
 | `LISTEN_EDITS` | `true` | B | Apply text edits in real-time |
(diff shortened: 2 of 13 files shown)


GeiserX:
Ran a 7-perspective review panel (logic, scope-injection architecture, ACL security, frontend, tests, concurrency/perf, simplicity) with adversarial verification, and addressed CodeRabbit — all in 3195384.

**Fixed**
- **FOLLOW_CHAT_MIGRATIONS worked in include-list/type mode but was dead in whitelist mode** — `_should_process_chat`'s whitelist branch only honored `CHAT_IDS`, so an adopted supergroup was never tracked live there. Now a dedicated followed set is consulted in that branch (whitelist strictness preserved). Didn't affect the reporter (include-list mode), but correct to fix.
- **Spurious migration warning** for all-groups-mode users whose new supergroup is already type-in-scope — suppressed via `should_backup_chat`; whitelist-mode followed fetch now honors exclude lists.
- **ACL membership probe now fails closed** on a DB error (deny → 403, log the exception class only) instead of surfacing a 500 (CodeRabbit Major), is cached with a short TTL, and avatar responses send `Cache-Control`; the sender-avatar `` is `loading="lazy"`; the initials monogram mirrors the visible sender-name fallbacks instead of `?`; the probe docstring corrected (CodeRabbit Minor).
- **Test-coverage gaps closed** (the two findings the verifier downgraded to low): a followed id is now proven to land in the swept dialogs across all three injection sites; a real-config listener-union test (the prior one used a MagicMock config that would pass even if the union line were deleted); a direct membership-probe test, the ACL allow/block endpoint path, `sender_avatar_url` wiring, and the fail-closed path; mock configs pin `should_skip_topic` (CodeRabbit Major).

**Declined with reasoning**
- CodeRabbit flagged `except ValueError, TypeError:` as a Critical Python-2-style clause — it's valid **PEP 758** on Python 3.14, which this project targets everywhere (CI runs 3.14, the Dockerfile builds 3.14, and `ruff format --target py314` *enforces* the comma form; parenthesizing fails `ruff format --check`). CodeRabbit's own `ast.parse` verification passed. No change.
- The migration-marker `LIKE` scan (flagged high by one reviewer) runs once per ~6h and is dwarfed by the sweep's own full-table work; a real optimization needs a `raw_data` index = schema change, deliberately out of scope here.

2344 tests pass (62 new across both issues); ruff clean.

PR #283, opened 2026-08-15
fix: stop one unreadable attachment from freezing a chat's backup forever

GeiserX:
## The problem

If Telegram ever hands back an attachment it can no longer retrieve, that chat stops being backed up — permanently — and the log still says the backup succeeded.

Telegram represents an unretrievable document as `documentEmpty`. Telethon turns that into `DocumentEmpty`, which is truthy but has no `.attributes`, so the media probe raised `AttributeError` while walking it.

What that cost:

- **In the scheduled sweep**, the exception aborted the entire dialog. The sync cursor had last been checkpointed *before* the offending message, so the next run resumed at the same message and failed identically. Every message newer than it was never archived, on any future run.
- **In the real-time listener**, the classification runs one line before the insert, so the message was dropped entirely — text included, with no row written.
- **Gap-fill cannot rescue it.** Gap detection only compares rows that already exist, so a missing tail is structurally invisible, and the gap-fill path calls the same failing code anyway.
- The run still printed `Backup completed successfully!`, and the PII policy strips chat identity from the error line, so an operator seeing it cannot tell which chat is stuck.

## The fix

An unusable document reference is now treated exactly like a missing one, which is the behaviour this code already documents for `document=None`.

Worth noting: simply guarding the attribute walk is *not* enough. With an empty iterable the loop body never runs and control falls through to `return "document"`, which would classify an expired attachment as a real one and send it down the download path. The document has to be recognised as unusable and returned as such.

The same unguarded walk existed in **four** other places — both `_get_media_filename` copies and the custom-emoji resolver. All of them now use the defensive form `message_utils.py` was already using; that file had the correct pattern all along and it had never been mirrored into the others.

## Verification

- New regression test asserts the classification in both capture paths and that neither filename builder raises.
- **Positive control:** reverting the guard reproduces the original failure exactly — `AttributeError: 'DocumentEmpty' object has no attribute 'attributes'` at `src/listener.py:682` — and restoring it goes green. The test was watched failing before it was trusted.
- Full suite: **2731 passed**, 171 subtests, 0 failures (baseline was 2725).
- `ruff check` and `ruff format --check` both clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of incomplete or expired media documents during backup and message capture.
  * Prevented errors when document metadata is missing.
  * Ensured unsupported media is skipped safely while valid documents continue to be recognized.
  * Preserved safe filename handling for documents without metadata.

* **Tests**
  * Added coverage for expired documents and standard document processing.

### DIFF
--- a/src/listener.py
+++ b/src/listener.py
@@ -678,8 +678,14 @@ class TelegramListener:
         elif isinstance(media, MessageMediaDocument):
             # Check document attributes to determine specific type
             if hasattr(media, "document") and media.document:
+                # DocumentEmpty is truthy but carries no .attributes at all. Its reference
+                # is unusable, so treat it exactly like a missing document rather than
+                # classifying it as a real one and sending it down the download path.
+                attributes = getattr(media.document, "attributes", None)
+                if attributes is None:
+                    return None
                 is_animated = False
-                for attr in media.document.attributes:
+                for attr in attributes:
                     attr_type = type(attr).__name__
                     if "Animated" in attr_type:
                         is_animated = True
@@ -712,7 +718,7 @@ class TelegramListener:
         if hasattr(message.media, "document") and message.media.document:
             doc = message.media.document
             mime_type = getattr(doc, "mime_type", None)
-            for attr in doc.attributes:
+            for attr in getattr(doc, "attributes", None) or ():
                 if hasattr(attr, "file_name") and attr.file_name:
                     original_name = attr.file_name
                     break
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai Declining the return-type annotations on the test methods, deliberately.

The guideline is real, but the established convention in this suite is the opposite: 2527 test methods have no return annotation against 194 that do, and `setUp` is unannotated in 107 of 108 cases. Matching the surrounding code matters more here than a rule applied to one new file, and annotating only this file would make it the odd one out.

Happy to revisit it as a suite-wide change if we want the guideline enforced consistently — that is a separate piece of work, not something to land inside a critical data-loss fix.

PR #286, opened 2026-08-15
fix: contain per-message capture failures, media symlinks, and archiver thumbnail memory

GeiserX:
## The problem

Three capture-side reliability gaps from the full audit, all of the "correct on the path you tested, missing on the mirror" kind.

**One bad message could still wedge a chat.** The `DocumentEmpty` crash was fixed in #283, but the deeper shape remained: any single message that failed to process aborted the *whole dialog*, and because the sync cursor was checkpointed before it, the next run resumed at the same message and failed again. That chat never advanced. The same failures leaked identifiers — several peer-scoped Telethon calls interpolate their exception text into the log, and Telethon spells the chat/user id out in that text.

**Shared-media symlinks could be silently broken.** Concurrent ingest published a transient flat name into the dedup store, so a chat symlink could point at a name that no longer existed — an archived attachment you can never open again. The sharding migration moved relative symlinks a directory deeper without rewriting their targets, breaking every one it touched, and a single unreadable file aborted the migration and with it container startup.

**A decompression bomb hit the archiver first.** The pre-generated thumbnail path had only a byte-size gate, so a ~300 KB / 96 MP image ballooned memory *inside the backup process* — measured at ~367 MB — and the thumbnail was written non-atomically.

## The fix

- **Per-message isolation:** a failed message is contained and counted, and `running_max_id` freezes at the first failure so the cursor can never checkpoint past unprocessed work — it stays retryable instead of being skipped forever. Topic-skips still advance normally.
- **PII:** peer-scoped call sites log the exception *type*, not the interpolated message.
- Terminal auth errors fail fast; the listener's flood nesting matches the backup path; no orphaned `.part` files; the listener detaches handlers on stop so a restart leaves one listener, not N+1.
- Atomic shared-media publish; symlink targets rewritten during re-sharding; per-file migration errors contained.
- Archiver thumbnails gated on decoded pixels (format-aware, so valid JPEGs still work) and written via atomic replace.

## Verification

- Full suite: **2763 passed**, 171 subtests, 0 failures (baseline 2725).
- Every new test proven able to go red before being trusted — including the cursor-freeze test (the cursor sails past the poison message without it) and the PII tests (the raw id appears in the log line without it).
- The archiver bomb re-run against an **unpatched shadow tree**: thumbnail written with ~367 MB RSS there, refused here.
- `ruff` check and format clean under the pinned version.

### DIFF
--- a/src/listener.py
+++ b/src/listener.py
@@ -285,6 +285,13 @@ class TelegramListener:
         # Real-time notifier for viewer WebSocket updates
         self._notifier: RealtimeNotifier | None = None
 
+        # Callbacks handed to client.add_event_handler, kept so stop() can detach
+        # them again. Telethon only ever appends, and the scheduler builds a NEW
+        # listener on the SAME shared client after every network blip, so without
+        # this every restart left another live instance dispatching the same
+        # updates: duplicate writes, duplicate broadcasts, duplicate downloads.
+        self._registered_handlers: list = []
+
         # Statistics
         self.stats = {
             "edits_received": 0,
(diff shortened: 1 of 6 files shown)


GeiserX:
@coderabbitai Both findings accepted as real. Taking them differently:

**Cursor freeze with no exit (`telegram_backup.py:1783`) — agreed, and it is the more important of the two.** You are right that `cursor_frozen` never clears, so a permanently unprocessable message makes every later run re-fetch and re-commit that chat's whole tail. I am merging anyway, deliberately: the behaviour this replaces is strictly worse — the dialog aborted entirely and every message newer than the bad one was *never archived*, on any run. Post-change the data is captured and the re-commits are idempotent upserts, so the cost is wasted work and a stalled `sync_status`, not data loss or corruption.

It needs a bounded exit and is filed as P1 with the analysis. Two constraints worth recording for whoever picks it up: gap-fill cannot serve as the safety net as-is, because `detect_message_gaps` only surfaces holes larger than the threshold (50) and a single skipped message is invisible to it; and a cross-run attempt counter needs durable state, which means a schema change rather than a patch-level fix.

**Rollback-safety in `migrate_shared_media.py` — agreed in principle, not fixed here.** The relink-before-relocate ordering can leave chat symlinks pointing at a destination that never materialised. Scoping it out of this PR because a correct fix is a transactional relocate with restore-on-failure, which is a heavier change than the containment this PR was for, and the current change is still a strict improvement (it fixes links that re-sharding previously broke outright, and stops one unreadable file aborting the whole migration). Filed alongside the cursor item.

PR #290, opened 2026-08-15
fix: close the viewer's websocket and thumbnail ACL bypasses

GeiserX:
## The problem

Four ways the viewer handed out data it should not have. All were reproduced with working exploits before being fixed, and re-run afterwards to confirm they are dead.

**Anyone could open a websocket with no credentials.** When `AUTH_PROXY_HEADER` is set together with `VIEWER_USERNAME`/`VIEWER_PASSWORD` — a combination the project documents and tests — the proxy branch swallowed the request. With no header its only rejection was `not AUTH_ENABLED and not ALLOW_ANONYMOUS_VIEWER`, which is false when auth *is* enabled, so the cookie check below it was unreachable and the socket connected with `allowed_chat_ids=None`: no ACL at all. It then received every live message, edit, deletion and reaction, message text included.

There is a second impact that needs no attacker: in that same configuration an ordinary logged-in viewer restricted to chats `{1,2}` also got a null ACL, so it could subscribe to any chat id. That is horizontal privilege escalation against the per-viewer ACL, reachable by simply logging in.

**A restricted viewer could read other chats' media.** `GET /media/thumb/200/<allowed>/%2e%2e/<forbidden>/<file>` — uvicorn unquotes before routing, so `%2e%2e` arrives as a real `..`. The ACL was applied to the raw string's first segment while the file was resolved from the joined path, so the two disagreed. `serve_media` already had the guard; the thumbnail route never got it.

**Any attachment could run JavaScript in the viewer's origin.** Archived `.html`/`.svg`/`.xhtml` documents were served inline, so opening one executed it as a same-origin document holding the viewer's session — stored XSS from a file any Telegram contact can send you.

**Archive paths reached the logs.** An unhandled error logged the request path, which carries the chat id and the sender's original filename. Redacting the app's own log line was not enough: Starlette's error middleware re-raises after calling the handler, and the server then logs the full traceback unconditionally, so a thumbnail failure still printed the ffmpeg argv — chat id and filename — into the container log.

## The fix

- Both transports resolve their principal through **one shared resolver** (proxy header, then cookie, then anonymous), so HTTP and websocket cannot drift apart again.
- Both media routes **check and normalise the path once, up front**, so the string that is authorised is the string that selects the file.
- Only image, video, audio and PDF stay inline; everything else is served as an attachment with `nosniff`.
- The redaction moved into middleware registered so it sits **inside** Starlette's error middleware — it responds without re-raising, so the exception never reaches the server's traceback logging. Status codes and JSON bodies are byte-identical to before.
- Access-controlled media is no longer marked cacheable by shared proxies; `no_download` viewers are not handed thumbnail URLs that always 403.
- Password, token and share-token hashing moved off the event loop, along with the blocking per-id avatar directory scan.
- Broadcasts iterate a snapshot, so a client connecting mid-send no longer aborts delivery to everyone else.

## Verification

- Every exploit re-run against **two real uvicorn servers** (patched vs unpatched) — the credential-less socket, the restricted viewer's ACL, the encoded `..`, and the log leak. Each confirmed live on the unpatched side and dead on the patched side.
- The log fix was verified at real-uvicorn fidelity: with the middleware stripped, stderr shows `Exception in ASGI application` and the full ffmpeg argv; with it in place, the chat id, filename, argv and that log line are all absent.
- Every new test proven able to go red before being trusted. One false green was caught and corrected during this work — a stale `__pycache__` made two tests look passing when they were not.
- Full suite: **2763 passed**, 171 subtests, 0 failures. `ruff` check and format clean.

## Behaviour change worth knowing

Clicking a non-media attachment (`.html`, `.svg`, `.txt`, `.zip`) now **downloads** it instead of rendering it in a tab. That is the point of the fix: rendering it made it a same-origin document holding your session.

## Known, pre-existing, not fixed here

The server's **access log** still records the full request line, which contains the chat id and filename for media requests. That predates this change and needs a logging-config fix rather than an application one; it is tracked separately.

## Summary by CodeRabbit

- **Bug Fixes**
  - Strengthened access controls for media, thumbnails, chat, and WebSocket connections.
  - Blocked unsafe path traversal and prevented potentially dangerous media from rendering inline.
  - Improved privacy by hiding download links from users without download access and protecting cached private media.
  - Improved authentication reliability across web and real-time connections.
  - Prevented sensitive details from appearing in error logs.
  - Improved stability during simultaneous connection and broadcast changes.
- **Tests**
  - Added comprehensive coverage for security, privacy, authentication, and connection stability improvements.

### DIFF
--- a/src/web/main.py
+++ b/src/web/main.py
@@ -7,13 +7,13 @@ v5.0: WebSocket support for real-time updates and notifications.
 """
 
 import asyncio
-import glob
 import hashlib
 import json
 import logging
 import os
 import secrets
 import time
+import traceback
 from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
@@ -28,6 +28,7 @@ from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import FileResponse, HTMLResponse, JSONResponse, StreamingResponse
 from fastapi.staticfiles import StaticFiles
 from sqlalchemy.exc import DBAPIError, OperationalError
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 from ..config import Config
 from ..db import DatabaseAdapter, close_database, get_db_manager, init_database
@@ -89,7 +90,11 @@ class ConnectionManager:
     async def broadcast_to_chat(self, chat_id: int, message: dict):
         """Broadcast a message to all connections subscribed to a chat."""
         disconnected = []
-        for websocket, subscribed_chats in self.active_connections.items():
+        # Snapshot first: send_json suspends, and a connect/disconnect landing in
+        # another task during that await mutates active_connections. Iterating it
+        # live raises RuntimeError at the `for`, which aborts the whole broadcast
+        # and silently drops the event for every client not yet reached.
+        for websocket, subscribed_chats in list(self.active_connections.items()):
             allowed = self._allowed_chats.get(websocket)
             if allowed is not None and chat_id not in allowed:
                 continue
@@ -107,7 +112,8 @@ class ConnectionManager:
     async def broadcast_to_all(self, message: dict):
         """Broadcast a message to all connected clients."""
         disconnected = []
-        for websocket in self.active_connections:
+        # Same snapshot rule as broadcast_to_chat: the dict can change under us.
+        for websocket in list(self.active_connections):
             try:
                 await websocket.send_json(message)
             except Exception as e:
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai On the CodeQL `py/path-injection` alert at `src/web/main.py:1176` — assessed as a false positive, with the containment verified by hand rather than assumed:

1. `_checked_media_path()` rejects any absolute path or `..` segment before the string is used for anything (`main.py:1017`), and it is the *single* string used for both the ACL decision and the file read — which is the actual bug this PR fixes.
2. The candidate is then resolved with `Path.resolve(strict=True)` (`:1117`).
3. The resolved path is rejected unless `is_relative_to(_media_root)` (`:1135`), so a symlink cannot escape either.

CodeQL does not model `is_relative_to` as a sanitizer, so the taint appears to survive. The same rule already fires on three other pre-existing lines in this file on `main`; this alert is that class following the refactor, not a new sink.

Not dismissing the alert in the UI — the JavaScript scanning added in #285 turned up two genuine SPA findings on `main` (a high-severity double-escaping issue in the message link builder, and a request-forgery pattern on interpolated fetch URLs), and both are filed for proper investigation rather than waved through. Keeping the alert list honest matters more than a green badge.

PR #291, opened 2026-08-15
fix: make a link in a message navigate to the host it displays

GeiserX:
## The problem

A link in an archived message could **display one host and navigate to another**.

`linkifyText` escaped the message text, then decoded `&amp;` back to `&` inside the matched URL to build the `href`. The browser then decodes that href a *second* time — so a character reference typed into a Telegram message became a real character in the link target.

A message containing:

```
https://good.example&#x40;evil.example/
```

rendered as a link reading `https://good.example…` while resolving to host **`evil.example`**, because `&#x40;` became `@` and demoted the visible host into userinfo. Anyone who can message the archived account could plant one.

Verified in a real DOM (jsdom + parse5) across 36 payloads. `&#64;`, `&#X40;`, `&#0000064;`, `&commat;`, the semicolon-less legacy forms, and `&colon;`+`&commat;` all behave the same; `&#10;`/`&Tab;` similarly corrupt the host.

**It is not XSS, and the report says so.** A literal quote or angle bracket can never reach the markup — raw quotes survive `escapeHtml` but are then percent-encoded, and the `&amp;` decode can only ever produce `&`. The URL pattern only matches `http(s)`, so `javascript:`/`data:` is unreachable. Across all 36 payloads the output was always exactly one `<a>` carrying only `href`, `target`, `rel` — never an extra tag, never an `on*` attribute. So: real finding, honestly scoped — a phishing primitive, not code execution. `rel="noopener noreferrer"` means the damage is the navigation itself.

## The fix

The escape-then-decode round trip is **removed**, not patched. The raw text is split on the URL pattern and each piece is escaped exactly once for the context it lands in: text segments for text, and for each URL a separate attribute-escaped `href` and an HTML-escaped visible label. There is no decode step left to exploit — which is also what closes the CodeQL `js/double-escaping` alert that surfaced this.

## Verification

- 120 regression cases execute the **real helpers lifted verbatim from the shipped template** under node, then parse the emitted markup with the stdlib `HTMLParser`, asserting on genuinely parsed hrefs and link text — no new dependency.
- Pinned invariants: href matches the visible text, the navigated host equals the displayed host, no payload resolves to `evil.example`, output is one inert anchor, scheme stays `http(s)`.
- **Positive control:** restoring the old algorithm turns 27 assertions red across 5 test functions; a full revert to main turns all 120 red. The test cannot silently pass against the old code.
- Ordinary links are byte-identical: same spans detected, same visible text, same `target`/`rel`, same resolved href, diffed old-vs-new over all 36 payloads.
- Full suite: **2975 passed**, 171 subtests, 0 failures. `ruff` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link handling to prevent URL and HTML entity injection.
  * Preserved visible message text and displayed link destinations accurately.
  * Ensured HTTP links, query parameters, and host resolution continue working correctly.
  * Rendered messages without links as safely escaped plain text.
  * Prevented malicious link attributes from becoming active or redirecting elsewhere.

* **Tests**
  * Added regression coverage for link rendering, URL escaping, inert attributes, and empty or null messages.

### DIFF
--- a/src/web/templates/index.html
+++ b/src/web/templates/index.html
@@ -5612,23 +5612,33 @@
                     return div.innerHTML
                 }
 
+                // Map a RAW url into href="..." in ONE pass: & escaped for attribute
+                // context, and the characters that must never sit raw inside the
+                // attribute percent-encoded. One pass is the whole point. The previous
+                // shape escaped the entire message and then decoded &amp; back to & while
+                // building the href, so message text could smuggle an entity through that
+                // second decode: `https://good.example&#x40;evil.example/` displayed
+                // good.example but the parser turned &#x40; into @ and the link navigated
+                // to evil.example.
+                const escapeUrlForAttr = (url) => {
+                    const replacements = { '&': '&amp;', '<': '%3C', '>': '%3E', '"': '%22', "'": '%27' }
+                    return [...url].map((char) => replacements[char] || char).join('')
+                }
+
                 const linkifyText = (text) => {
                     if (!text) return ''
-                    const escaped = escapeHtml(text)
-                    const urlRegex = /(https?:\/\/[^\s<]+)/g
-                    return escaped.replace(urlRegex, (match) => {
-                        // Build a safe href: decode HTML entities back to raw URL,
-                        // then percent-encode characters that break href="..." context.
-                        // Note: escapeHtml (textContent/innerHTML) does NOT encode " or '
-                        // so we must handle raw quotes directly, not just entities.
-                        const href = match
-                            .replace(/&amp;/g, '&')
-                            .replace(/&lt;/g, '%3C')
-                            .replace(/&gt;/g, '%3E')
-                            .replace(/"/g, '%22')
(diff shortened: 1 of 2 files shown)


GeiserX:
On the CodeQL `py/incomplete-url-substring-sanitization` alert at `tests/test_linkify_escaping.py` — leaving it as-is, deliberately.

The flagged line is `assert link_text.startswith("https://good.example")`. That is a test assertion about what the link **visibly reads as**, not URL validation: the whole point of the test is that the text reads as one host while the href resolves to another, and the very next line asserts `urlsplit(href).hostname != "evil.example"` on the parsed href. A prefix check is the correct expression of "a human reading this sees good.example".

Rewriting it to parse the display text would make it wrong rather than safer — the payload deliberately puts a character reference inside the host, so parsing the visible text does not yield `good.example`.

CodeQL is not a required check here and the alert is confined to test code, so it is not being suppressed in the UI either — the alert list stays honest rather than groomed for a green badge.

PR #292, opened 2026-08-15
fix: give the frozen sync cursor a way out

GeiserX:
## The problem

Per-message isolation (#286) holds the sync cursor behind the first message that fails to process, so a transient failure is retried next run instead of being silently skipped. That part was right — but **the state had no exit**.

A message that can never be processed pinned the cursor forever, and every later run re-fetched and re-committed that chat's entire tail. Measured over six consecutive runs of one chat with a permanently failing message, adding two new messages between runs:

| run | messages fetched | cursor after | rows written (cumulative) |
|----|----|----|----|
| 1 | 10 | 4 | 9 |
| 2 | 8 | 4 | 16 |
| 3 | 10 | 4 | 25 |
| 4 | 12 | 4 | 36 |
| 5 | 14 | 4 | 49 |
| 6 | 16 | 4 | 64 |

The work per run grows without bound while the cursor never moves, and `sync_status` permanently misreports progress. No corruption — the re-commits are idempotent upserts — but unbounded waste and a stalled cursor.

## The fix

A message is passed over once it has failed on **two separate runs**, and the ids passed over are recorded durably so the skip is never silent. The first failure behaves exactly as before, so a transient error still gets its retry — that guarantee is pinned by its own test.

**No schema change.** The record lives in the existing metadata table, one row per chat that has ever had a failure. Reading it is lazy, so a dialog that processes cleanly makes zero extra database calls. Every failure mode of that record — missing, malformed, or a database that cannot store it — degrades to the previous freeze behaviour rather than to an early skip.

Deliberately no migration: two workstreams are rewriting the schema for 8.0 right now, and a migration from here would collide with both.

## Verification

- **The reviewer built an independent harness from scratch** rather than reusing the implementation's tests, and drove both a pristine pre-fix tree and the patched one.
- Unpatched: 304 fetches over 15 runs, per-run cost growing, cursor pinned forever. Patched: **44 fetches, constant 2 per run from run 3 on** — a one-time overhead of 6 over a clean baseline, not a growing one.
- **Positive control:** three separate reverts, each disabling one part of the fix; the whole-exit revert produced `AssertionError: 4 != 10` — the cursor stuck where it used to stick — across 11 failing tests.
- Full suite: **2872 passed**, 171 subtests, 0 failures. `ruff` clean.

Operator-visible: a warning reports how many messages were passed over and that their ids are in the backup metadata — counts only, no ids in logs, per the project's PII rule.

## Summary by CodeRabbit

- **Bug Fixes**
  - Added bounded retries for messages that repeatedly fail during backup.
  - Failed messages are retried across runs before being skipped, preventing the sync process from stalling indefinitely.
  - Progress is now saved when a message is ultimately skipped, preventing silent data omissions.
  - Added durable tracking and capped history for skipped messages.
  - Improved handling of temporary failures and incomplete metadata.

### DIFF
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -122,6 +122,16 @@ MEDIA_REFRESH_TIMEOUT_SECONDS = _get_int_env("MEDIA_REFRESH_TIMEOUT_SECONDS", 12
 # Hard wall-clock bound on the #234 whitelist dialog scan — the count limit alone
 # cannot prevent a wedged-connection hang, which is what #95 was about.
 WHITELIST_RESOLVE_SWEEP_TIMEOUT_SECONDS = 300
+# Bounded retry for a message _process_message cannot handle (#286 follow-up).
+# Holding the sync cursor behind a failure is right for a TRANSIENT one, but a
+# PERMANENT one would hold it there forever: every later run would re-fetch and
+# re-commit the entire tail of that chat — a window that keeps growing — and the
+# chat's recorded progress would never advance again. After this many separate
+# runs have failed on the SAME message the cursor is allowed past it.
+MESSAGE_MAX_PROCESS_ATTEMPTS = 2
+# Cap on the ids kept in a chat's give-up record so it cannot grow without bound.
+# The running total is stored separately and stays exact.
+MESSAGE_GIVE_UP_RECORD_LIMIT = 500
 
 
 def _media_retry_backoff_seconds(attempt: int) -> float:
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai Accepted and fixed in 6aeef74 — this was a real hole, and exactly the one the mechanism exists to close.

You are right that the give-up branch only mutated the in-memory record. A batch checkpoint later in the same run can commit a cursor past the passed-over id, so a process dying in between left the cursor beyond a message nothing remembered skipping: never retried, never reported.

The write now happens inside the give-up branch before the cursor is allowed to advance, as you proposed. A give-up is rare, so the normal path is unaffected.

Added `TestTheRecordSurvivesACrashMidDialog`, which drives a run that dies mid-dialog after the give-up and asserts the id is already on disk. **Positive control:** removing the new write turns `test_given_up_id_is_on_disk_even_if_the_run_dies_after_it` red; restoring it gives 19 passed. Being precise about what that control proves — the companion assertion in that class passes either way, because the cursor does not always reach past the poison id in the crash run; the first test is the load-bearing one.

Full suite 2874 passed, 0 failures, ruff clean.

PR #297, opened 2026-08-15
fix: never advance the cursor past a message whose skip record failed to persist

GeiserX:
## The problem

#292 made the backup record every message it gives up on, so the cursor never silently passes an unarchived message. The pre-8.0 re-review found the one crack left in that promise: the record write itself could fail — `set_metadata` has no `database is locked` retry, and the listener writes the same SQLite file concurrently — and the failure was swallowed at debug level while the cursor advanced anyway. The run's WARNING then claimed "their ids are kept in the backup metadata" when nothing had landed. Reproduced end to end before fixing: cursor at 10, message 5 never committed, record still showing the old freeze.

## The fix

- `set_metadata` / `get_metadata` gain `@retry_on_locked()`, mirroring `update_sync_status` — transient lock contention no longer reaches the swallow at all.
- `_save_message_failures` now reports whether the record landed, and the give-up branch advances the cursor **only when it did**. A failed write holds the freeze for one more run, which converges; the accurate "cursor stays behind them" warning covers the held case, and the "ids are kept" warning can now only fire for records that actually landed.
- Housekeeping from the same review: the `frozen_id`/`runs` half of the record is cleared once the frozen message finally succeeds (free in-memory check when the record was already loaded, plus one gated read at dialog end), so it no longer survives forever after recovery. A dedicated guard test pins that this clearing can never touch the give-up half.

## Verification

- Five new tests in `tests/test_cursor_freeze_exit.py`; the four behavioural ones were watched fail against the pre-fix code with exactly the reported symptoms (cursor passed with no record; false WARNING; lock errors reaching the swallow; stale frozen half).
- The lock-retry test drives the real decorated `set_metadata` through two genuine-shaped `database is locked` errors and asserts the give-up landed durably.
- Independent reviewer re-ran the original finding's repro: not reproducible here, still reproducible on unfixed `main` — and the full suite with a real PostgreSQL leg: **3042 passed, 0 failed, 0 skipped**. ruff check and format clean.

No version bump: still 7.33.6.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when message-failure records cannot be saved.
  * Prevented cursor progress until skipped-message records are successfully persisted.
  * Added automatic retry handling for temporary database locks.
  * Cleared stale frozen-message states after successful recovery.
  * Improved warnings to distinguish held failures from recorded skips.
  * Ensured failed message handling can resume safely on a later run.

### DIFF
--- a/src/db/adapter.py
+++ b/src/db/adapter.py
@@ -195,9 +195,12 @@ def retry_on_locked(
 
                     last_exception = e
                     if attempt < max_retries:
+                        # Type name only: the raw exception text can carry the SQL
+                        # statement, bound values, or a connection DSN, and this
+                        # wraps writers whose payloads identify chats.
                         logger.warning(
                             f"Database error on {func.__name__}, attempt {attempt + 1}/{max_retries + 1}. "
-                            f"Retrying in {delay:.2f}s... Error: {e}"
+                            f"Retrying in {delay:.2f}s... Error type: {type(e).__name__}"
                         )
                         await asyncio.sleep(delay)
                         delay = min(delay * backoff_factor, max_delay)
--- a/src/db/adapter.py
+++ b/src/db/adapter.py
@@ -523,6 +523,7 @@ class DatabaseAdapter:
 
     # ========== Metadata Operations ==========
 
+    @retry_on_locked()
     async def set_metadata(self, key: str, value: str) -> None:
         """Set a metadata key-value pair."""
         async with self.db_manager.async_session_factory() as session:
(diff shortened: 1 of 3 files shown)


GeiserX:
The outside-diff-range finding on `retry_on_locked` is addressed in 84c6510: the retry warning now logs `type(e).__name__` only — the raw exception text can carry the SQL statement, bound values, or a connection DSN, and the decorator now wraps writers whose payloads identify chats. Detection still inspects the string; only the log line changed.

PR #300, opened 2026-08-15
fix: harden migration 021 - audit indexes, fail-closed token backfill, crash-loop guard

GeiserX:
## The problem

Three lanes of the pre-8.0 re-review independently converged on gaps in migration 021 — all safe to amend in place because 021 is merged but unreleased (every real user upgrades from ≤7.33.6):

1. **The audit-log indexes never materialize for the most common install.** A create_all-provisioned 7.x SQLite archive upgraded through the shipped entrypoint keeps `viewer_audit_log` unindexed forever — measured 22.3ms vs 3.1ms on the admin audit page at 500k rows — and the parity gate can't see it because it only compares freshly built schemas.
2. **The `is_revoked NULL→0` backfill resurrects dead tokens.** The validator filters `is_revoked == 0`, which excludes NULL — so a NULL token was unusable in 7.x, and backfilling 0 brings it back to life. Security-relevant direction.
3. **Crash atomicity on SQLite was an emergent property of statement order.** If batch DDL ever runs first (a manually repaired database), pysqlite autocommits `CREATE TABLE _alembic_tmp_<t>`; a crash there leaves a poison table that crash-loops every subsequent start, with no cleanup path anywhere.

## The fix

1. Guarded `op.create_index` for `idx_audit_log_username`/`idx_audit_log_created` on both backends, mirroring migration 007's style.
2. The backfill flips to `NULL→1`: a token in unknown state stays dead. Every other backfilled column was re-audited: `no_download` is the only other access gate and is read via Python truthiness where NULL and 0 behave identically; the `Media.downloaded` fill re-queues a formerly invisible row (safe, work-redoing direction) — the docstring now states this precisely instead of a false absolute.
3. `DROP TABLE IF EXISTS "_alembic_tmp_<t>"` before every SQLite batch rebuild, upgrade and downgrade — self-healing against the poison table. The orphan-reaction DELETE was deliberately **not** made unconditional: unguarded it would delete rows on create_all archives, contradicting the documented we-do-not-delete-data-we-are-not-forced-to decision.
4. The docstring now names both deliberate leftovers (the unnamed reactions→users FK, and pre-existing orphan reaction rows on create_all archives).

## Verification

New `tests/test_migration_021_hardening.py`, both backends: index creation, NULL-token stays dead, poison-tmp-table self-heal, and re-run no-op. The three behavioural cases were watched fail against the pre-fix migration — including the exact `table _alembic_tmp_media already exists` crash-loop. `tests/test_schema_parity.py` stays green with an **empty** allow-list on both backends. Independent reviewer re-ran the probes both ways. No version bump: still 7.33.6.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved confidence in database upgrades by covering missing audit indexes, revoked-token handling, and cleanup of temporary migration tables.
  * Ensured repeated application of migration 021 remains safe and does not alter already-correct schemas.

* **Tests**
  * Added coverage for migration behavior across SQLite and PostgreSQL, including legacy schemas and interrupted rebuild scenarios.

### DIFF
--- a/alembic/versions/20260815_021_align_schema_with_orm.py
+++ b/alembic/versions/20260815_021_align_schema_with_orm.py
@@ -14,7 +14,15 @@ What it changes, and why each is safe:
   migrations never said so, so a PostgreSQL install accepted NULLs that the
   SQLite install of the same version rejected. Every column is backfilled with
   the ORM's own default *before* it is tightened, so a database holding NULLs
-  converges instead of crash-looping on ``SET NOT NULL``.
+  converges instead of crash-looping on ``SET NOT NULL``. The one fill that is
+  not the ORM default is ``viewer_tokens.is_revoked``, filled with ``1``: the
+  token validator only reads rows with ``is_revoked = 0``, so a NULL there was
+  an unusable token, and a token in an unknown state must stay unusable. It is
+  alone in that. The only other backfilled gate, ``no_download``, is read for
+  Python truthiness, where NULL and ``0`` are the same answer; the remaining
+  fills are timestamps, cursors and display flags whose worst case is redoing
+  idempotent work (a zeroed ``media.downloaded`` becomes a download candidate
+  again).
 * **Server defaults.** ``created_at``/``updated_at`` gain the ``func.now()``
   default the ORM declares. ``messages.is_pinned``/``is_outgoing`` and
   ``media.downloaded`` lose the literal ``0`` the ORM does not declare (their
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

GeiserX:
On the annotations nitpick: declining for consistency — this file deliberately follows the dual-backend migration-test family it belongs to (`test_schema_parity.py`, `test_migration_022.py`, `test_createall_race.py`), which is module-level pytest with fixture parameters and no signature annotations, because it leans on pytest machinery (`tmp_path`, dual-backend fixtures, parametrize) that doesn't compose with the annotated-unittest style of the older suites. Keeping the family uniform beats annotating one member.

Issue #295, opened 2026-08-15
[Bug]: get_dialogs() never completes for accounts with enough dialogs to trip FloodWait (retry restarts pagination from scratch)


jordanfelle:
### Version
v7.7.0 (reproduced; confirmed the relevant `_get_dialogs()` / `call_with_flood_retry()` logic is unchanged on `main` as of this writing, v7.33.6)

### Deployment Method
Docker Compose

### Database Type
PostgreSQL

### Bug Description

On an account with a large number of dialogs (~1,900 in this case), the initial (non-whitelist) backup never completes `get_dialogs()` — it fails with the same ~25s `FloodWaitError` on every single scheduled run, no matter how far apart the runs are spaced (tested 6h apart, 4 runs, identical failure every time) or how many retries `MAX_FLOOD_RETRIES` allows (tested up to 40 consecutive retries — every one hit an essentially identical wait, no decay or jitter that would let it through).

**Root cause:** the client runs with `flood_sleep_threshold=0` (#124), so `client.get_dialogs()` — which paginates internally in ~100-dialog chunks via repeated `GetDialogsRequest` calls — aborts its *entire* call the moment any single internal page hits a FloodWait. `call_with_flood_retry` then sleeps and calls `get_dialogs()` again *from scratch*, re-walking the same already-successful early pages before hitting the same later page's FloodWait again. For an account whose dialog count requires enough pages to reliably trip a wait, this never converges — it's not cumulative account-level throttling wearing off over time (which more retries or a longer backoff would fix), it's the same deterministic wall being re-hit on every attempt.

This is the same failure class as #232 (mid-download FloodWaits restarting a media transfer from byte 0), just for `get_dialogs()` instead of media downloads. #232 was fixed via `MEDIA_FLOOD_SLEEP_THRESHOLD` / `absorb_media_floods()`, which lets Telethon absorb short floods *in place* (sleep + resume) instead of raising all the way out. No equivalent exists for `get_dialogs()`.

Whitelist mode (`CHAT_IDS`) sidesteps this entirely by skipping `get_dialogs()` (#95/#96), but that's not a fix for accounts that intentionally back up everything via `CHAT_TYPES`.

### Expected Behavior

The initial dialog listing should be able to complete for large accounts, the same way media downloads now survive floods mid-transfer instead of endlessly restarting.

### Steps to Reproduce

[...]

PR #301, opened 2026-08-15
perf(db): add a pg_trgm GIN index for the viewer's message search

jordanfelle:
## Summary

- Message search (`Message.text.ilike('%term%')`, used by both the viewer's chat search and message search) currently forces a **full sequential scan** of the `messages` table on every query - a leading-wildcard `ILIKE` can't use a plain B-tree index at all. Cost scales linearly with archive size.
- Adds a `pg_trgm` GIN index (`idx_messages_text_trgm`) on `messages.text`. Verified live on a ~100k-row PostgreSQL instance:

  | | Plan | Rows filtered | Execution time |
  |---|---|---|---|
  | **Before** | `Seq Scan on messages` | 6,080 removed by filter | **41.764 ms** |
  | **After** | `Bitmap Heap Scan` using `idx_messages_text_trgm` | 6 removed by index recheck | **0.438 ms** |

  **~95x faster** on this dataset - and the more important number is the shape of the cost curve, not the multiple: a sequential scan is O(table size), the trigram index is roughly O(matching rows). The seq scan gets slower every single day this project's own incremental backups add more messages; the index does not.

## Type of Change

- [ ] Bug fix
- [x] New feature (non-breaking - a new index, additive only)
- [ ] Breaking change
- [ ] Documentation update

## Database Changes

- [x] Schema changes (Alembic migration required) - `alembic/versions/20260815_022_add_message_text_trgm_index.py`
- [ ] No database changes

## Data Consistency Checklist

- [x] N/A - index-only change, no data written or read differently

## Testing

- [x] Tests pass locally (`python -m pytest tests/ -v`) - 3038 passed, 1 skipped, run with a real local PostgreSQL 18 server so the PostgreSQL leg of the parity/adapter suites actually executed (not just the SQLite leg)
- [x] Manually tested in development environment - `EXPLAIN ANALYZE` before/after on a live instance with real backed-up data (see Summary table); new catalog-level test (`test_trgm_index_exists_on_postgresql`) added so a future refactor can't silently drop the index type back to a plain B-tree without a test failing

## Security Checklist

- [x] No new input surface (search input handling is unchanged - same `ilike(..., escape="\\")` call as before, just now index-accelerated)
- [x] N/A - no auth-relevant code touched

## Deployment Notes

- SQLite deployments: no behavior change. `pg_trgm`/`gin_trgm_ops` are PostgreSQL-only; on SQLite the same-named index is still created (via `Base.metadata.create_all()`, since SQLAlchemy silently drops PostgreSQL-only dialect kwargs elsewhere) but as a plain, currently-unused B-tree, purely to keep `tests/test_schema_parity.py` passing (ORM- and Alembic-built schemas must agree on every dialect, not just PostgreSQL).
- PostgreSQL deployments: migration 022 runs `CREATE EXTENSION IF NOT EXISTS pg_trgm` (a standard, always-available contrib extension - doesn't require any server-level `shared_preload_libraries` change, unlike `pg_stat_statements`) before creating the index. No `CONCURRENTLY` - this project's other index migrations (e.g. #213/017) also run inside Alembic's normal transactional DDL, so this one does too for consistency; on an already-large `messages` table the index build will briefly hold a lock, same cost class as a one-time `REINDEX`.

## Analysis / why this approach

- **Why `pg_trgm` and not `tsvector`/full-text search:** the existing query is already a literal substring match (`ILIKE '%term%'`), which is what users searching chat history actually expect (Telegram's own client search is substring-ish, not stemmed/tokenized). `pg_trgm` accelerates exactly that predicate shape with zero query-code changes required - `tsvector` would need the query rewritten to `@@ to_tsquery(...)`, changing match semantics (stemming, no partial-word matches) as a side effect of a "just make it fast" fix.
- **Why the index also has to exist on SQLite:** discovered this the hard way - the ORM (`Base.metadata.create_all()`, authoritative for SQLite per #294/021's docstring) doesn't know to skip an `Index()` for a specific dialect; `postgresql_using`/`postgresql_ops` are dialect-scoped kwargs SQLAlchemy just drops elsewhere, so `create_all()` still emits a plain index on SQLite. Making the Alembic migration `return` early on non-PostgreSQL (my first attempt) skipped creating that same index there, and `test_orm_and_alembic_schemas_agree_on_sqlite` correctly caught the resulting drift (`index-only-in-orm`). The migration now creates the same plain index on every dialect and only adds the GIN/`pg_trgm_ops` treatment on PostgreSQL.
- **Why a catalog check instead of `EXPLAIN`:** first draft of the new test asserted on the query plan directly. On a near-empty per-test table the planner correctly prefers a sequential scan over any index (an index scan on 1-2 rows has *more* overhead, not less) - regardless of whether the trigram index exists, so that assertion would have been a size-dependent flake, not a real check of the fix. Querying `pg_index`/`pg_am`/`pg_opclass` for the actual access method and operator class is deterministic at any table size.

I have a production instance already running this (verified the before/after numbers above against it) - happy to answer questions about real-world behavior at whatever scale you want more data on.

## Summary by CodeRabbit

* **Performance**
  * Improved message text-search performance in PostgreSQL databases with optimized trigram indexing.
  * Automatically enables the required PostgreSQL text-search extension during database setup.
* **Compatibility**
  * SQLite databases continue to operate without PostgreSQL-specific text-search indexes.
* **Tests**
  * Added integration coverage for PostgreSQL index creation and SQLite compatibility.
  * Improved migration tests to remain reliable as the migration chain evolves.

### DIFF
--- a/alembic/versions/20260815_022_add_message_text_trgm_index.py
+++ b/alembic/versions/20260815_022_add_message_text_trgm_index.py
@@ -1,0 +1,91 @@
+"""Add a pg_trgm GIN index on messages.text for the viewer's search (PostgreSQL).
+
+get_messages_paginated / get_all_chats' search filters both use
+``Message.text.ILIKE('%term%')`` (leading wildcard - substring search, not a
+prefix match). A plain B-tree index can't serve that pattern at all, so every
+search was a sequential scan of the whole messages table, with cost growing
+linearly as the archive grows. Verified on a live ~100k-row instance:
+
+    EXPLAIN ANALYZE SELECT id FROM messages WHERE text ILIKE '%test%' LIMIT 20;
+
+    Before: Seq Scan on messages (actual time=4.556..36.063, Rows Removed by
+            Filter: 6080) - Execution Time: 41.764 ms
+    After:  Bitmap Heap Scan on messages using idx_messages_text_trgm
+            (actual time=0.147..0.368) - Execution Time: 0.438 ms
+
+~95x faster on this dataset, and critically the scan cost moves from O(table
+size) to roughly O(matching rows) - a sequential scan degrades linearly as
+more messages are backed up, while the trigram index does not.
+
+The index is created on every dialect (same name/column, so ORM- and
+Alembic-built schemas agree per tests/test_schema_parity.py), but only
+PostgreSQL gets the GIN/pg_trgm treatment that actually makes it useful.
+SQLite has no pg_trgm equivalent, so it gets a plain B-tree there instead -
+harmless, if unused for this particular query pattern, and matches what
+``postgresql_using``/``postgresql_ops`` (dialect-scoped kwargs on the
+models.py Index()) already produce via ``Base.metadata.create_all()``:
+SQLAlchemy silently drops PostgreSQL-only kwargs on every other dialect
+rather than erroring.
+
+No CONCURRENTLY: this project's existing index migrations run inside
+Alembic's normal transactional DDL (see #213/017), so this one does too for
+consistency. On an already-large `messages` table this will briefly hold a
+lock for the duration of the index build - acceptable for the same reason
+prior structural changes to this table were: migrations run once, and this
+one is asymptotically the same cost class as REINDEXing an existing index.
+
+Revision ID: 022
+Revises: 021
+Create Date: 2026-08-15
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "022"
+down_revision: str | None = "021"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+TABLE_NAME = "messages"
+INDEX_NAME = "idx_messages_text_trgm"
+
+
+def _index_exists(inspector: sa.Inspector) -> bool:
+    return INDEX_NAME in {ix["name"] for ix in inspector.get_indexes(TABLE_NAME)}
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    dialect = conn.dialect.name
+    is_postgresql = dialect == "postgresql"
+
+    if is_postgresql:
+        conn.execute(sa.text("CREATE EXTENSION IF NOT EXISTS pg_trgm"))
+
+    inspector = sa.inspect(conn)
+    # Idempotent: create_all()-provisioned databases may already have the index.
+    if TABLE_NAME in inspector.get_table_names() and not _index_exists(inspector):
+        # Same index name/column on every dialect (matches what models.py's
+        # Index() produces via create_all(), so ORM- and Alembic-built
+        # schemas agree - see tests/test_schema_parity.py). Only PostgreSQL
+        # gets the GIN/pg_trgm treatment that makes it actually fast;
+        # postgresql_using/postgresql_ops are dialect-scoped kwargs that
+        # SQLAlchemy silently drops on every other dialect, producing a
+        # harmless, unused plain B-tree there instead.
+        kwargs = {"postgresql_using": "gin", "postgresql_ops": {"text": "gin_trgm_ops"}} if is_postgresql else {}
+        op.create_index(INDEX_NAME, TABLE_NAME, ["text"], **kwargs)
+
+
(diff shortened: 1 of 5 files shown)


jordanfelle:
Thanks for the review! Checked the module-level suggestion against this file directly: `grep -n "^async def test_\|^def test_\|^class Test" tests/test_db_adapter_real_engine.py` shows every test in this file (all three classes: `TestUpdateSyncStatusRealEngine`, `TestMessageUpsertConflictRealEngine`, `TestPaginationRealEngine`) lives inside a class — there are no module-level tests here to be consistent with. `test_trgm_index_exists_on_postgresql` sits in `TestPaginationRealEngine` right next to `test_search_escapes_sql_wildcards`, the other search-behavior test it's a direct sibling of, which matches the file's actual (100% class-based) convention. Left it as-is.

GeiserX:
Reviewed in full — the approach is right, and this is another genuinely well-built contribution: the measured plan flip (seq scan → bitmap heap scan, with the honest note that the cost-curve shape matters more than the 95×), the extension handled on *both* provisioning paths (migration + a dialect-scoped `before_create` DDL hook for the `create_all` path), idempotent inspector guards, and a catalog-based test that deliberately avoids EXPLAIN-plan flakiness on small tables. Thank you, @jordanfelle.

Two things before it can merge, one of which you couldn't have known:

1. **Revision collision.** `revision = "022"` collides with #302, the 8.0 multi-account migration that also chains from 021 and is about to merge — two heads would break `alembic upgrade head` for everyone. Once #302 lands, this needs renumbering to `023` with `down_revision = "022"` (filename `20260816_023_add_message_text_trgm_index.py` per the repo convention) and a rebase — the `messages` PK becomes `(account_id, chat_id, id)` there, which is orthogonal to this index. Happy to do the renumber+rebase for you if you prefer ("Allow edits by maintainers" is on) — say the word.

2. **SQLite shouldn't pay for an index it can never use.** As written, SQLite gets a plain B-tree over the *entire* `messages.text` — and a B-tree on a text column duplicates that text into the index, so on a multi-gigabyte SQLite archive (the most common install) that's real, permanent file growth for zero query benefit. Suggest making it PostgreSQL-only on both paths: in `models.py`, `Index("idx_messages_text_trgm", "text", postgresql_using="gin", postgresql_ops={"text": "gin_trgm_ops"}).ddl_if(dialect="postgresql")` — verified available on the pinned SQLAlchemy 2.0.49, returns `Self` so it composes inline in `__table_args__` and `create_all` skips it on SQLite; in the migration, create the index only under `is_postgresql`. The schema-parity gate stays green because both provisioning paths then agree per backend (and `KNOWN_DIFFERENCES` stays empty, which we guard). Worth extending your catalog test with the SQLite-absence assertion so the split is pinned.

Minor: the `#295-perf` comment references should point at this PR — #295 is the FloodWait issue.

Timing: this rides the 8.0.0 release right after the multi-account core, with credit in the release notes. Your two contributions today have both been a pleasure to review.

GeiserX:
Went ahead with the maintainer edit as offered, since the 8.0 core (#302) just merged and this was the one thing between your index and the release train — hope that's alright. What changed on your branch (`d68c392`, rebased onto `main`):

- **Renumbered to revision 023** chaining from the multi-account 022 (the collision from my earlier comment), filename `20260816_023_add_message_text_trgm_index.py`.
- **PostgreSQL-only on both provisioning paths**: your `Index()` now carries `.ddl_if(dialect="postgresql")` and the migration creates only under the dialect guard — the SQLite same-name B-tree would have duplicated the whole `text` column into an index its ILIKE plan can never use. A SQLite-absence test pins the split next to your GIN catalog pin.
- Adapted the 022 suite's head assertions to read the live chain head (and scoped its crash matrix to its own revision), so the *next* migration after yours doesn't re-break them either.

Everything you wrote is otherwise untouched — the extension handling on both paths, the catalog-based test, the docstrings. Full suite on the rebased branch: **3192 passed, 0 failed** with a real PostgreSQL leg. Once CI agrees, this merges and ships in 8.0.0 with credit to you alongside #296. Note the branch was force-pushed (rebase + renumber required it) — `git fetch && git reset --hard origin/perf/pg-trgm-message-search` locally if you have WIP on it.

PR #306, opened 2026-08-16
fix: revocation now closes the live channels it revokes

GeiserX:
## The problem

8.0.0's headline is entitlements that fail closed — but revocation didn't reach the live channels. An open websocket outlived logout, viewer deactivation or deletion, token revocation and session expiry, because the delivery re-check reads the connect-time frozen context, never the session store. And push subscriptions had no revocation path at all: nothing deleted the rows on any trigger, delivery trusted the subscribe-time grant snapshot, and logout never unsubscribed the browser — so a revoked share token kept receiving notifications carrying sender names and 100 characters of message text. (Beads 9t6.4.6, 9t6.4.7, 9t6.4.10, 9t6.4.11, re-verified at file:line against current main before this fix.)

## The fix

- **Sockets carry revocation coordinates** (username, session key, token id, frozen at upgrade time from the same resolver that admitted them). `ConnectionManager.close_for(...)` closes matching sockets with the existing 4001 code, dropping their state before the close so a concurrent broadcast cannot deliver to a revoked socket.
- **Two choke points revoke everything.** `_invalidate_user_sessions` / `_invalidate_token_sessions` now kill sessions, close sockets, and purge the principal's push rows — covering viewer update/deactivate/delete and token revoke/update/delete. Logout does the same for its own session; the expiry sweep closes the sockets of sessions it expires, and the request-path expiry pops (which remove sessions the sweep would never see) close their own — the reverse-hole of the sweep-only design, found and tested.
- **Owner-liveness backstop in push delivery**: after the unchanged entitlement filter, rows whose owner no longer resolves (inactive/deleted viewer, revoked/expired token) are dropped — at most two queries total, never one per row; env-configured principals pass through since the database cannot prove them dead.
- **Browser hygiene**: `performLogout` unsubscribes from push and clears the auto-resubscribe flag before the cookie dies; the server purge remains the guarantee.

## Verification

- New `tests/test_revocation_closes_channels.py`: 12 scenarios × both backends on a real migrated database, socket revocation proven by watching the close frame arrive instead of the broadcast. Plus a node-harness test running the real extracted `performLogout`.
- **13 behavioural tests watched red against pre-fix code** — the six socket tests failed with the frame still delivered, and the backstop's red was strengthened by stubbing the new symbols as no-ops so it failed on behaviour, not a missing attribute. Two deliberate non-regression controls prove the filter is neither a global mute nor a replacement for the grant check.
- Full suite both backends: **3256 passed, 0 failed** with a real PostgreSQL leg (run twice: implementer and orchestrator independently). ruff check and format clean.

## Stated residuals

A fully idle expired socket persists up to one cleanup sweep (15 min) — the deliberate price of keeping session checks off the delivery hot path. The socket registry is per-process; the viewer image runs a single uvicorn worker, which preserves this by construction (adding `--workers` would silently make socket revocation per-worker — documented in code). The notification preview text itself (bead 9t6.4.10's lock-screen angle) is a product decision left untouched.

No version bump: still 7.33.6.

## Summary by CodeRabbit

* **New Features**
  * Real-time connections now close automatically when sessions expire, users are logged out or deactivated, or shared access is revoked.
  * Push notifications are delivered only to active, authorized accounts and valid shared sessions.
* **Bug Fixes**
  * Logging out now unsubscribes the current browser from push notifications and clears its local push settings.
  * Push subscriptions are removed when associated access is revoked or an account is deactivated or deleted.
* **Tests**
  * Added coverage for connection closure, subscription cleanup, and notification eligibility.

### DIFF
--- a/src/db/adapter.py
+++ b/src/db/adapter.py
@@ -34,6 +34,7 @@ from .models import (
     Message,
     MessageVersion,
     Metadata,
+    PushSubscription,
     Reaction,
     SyncStatus,
     User,
@@ -3760,6 +3761,20 @@ class DatabaseAdapter:
         """Delete all sessions created from a specific share token."""
         async with self.db_manager.async_session_factory() as session:
             result = await session.execute(delete(ViewerSession).where(ViewerSession.source_token_id == token_id))
+            await session.commit()
+            return result.rowcount
+
+    @retry_on_locked()
+    async def delete_push_subscriptions_for_username(self, *, username: str) -> int:
+        """Delete every push subscription owned by ``username``. Returns count deleted.
+
+        The revocation half of push ownership: a subscription is a delivery
+        channel that survives the session that created it, so every path that
+        invalidates a principal's sessions deletes its push rows through here.
+        Writes push_subscriptions and nothing else.
+        """
+        async with self.db_manager.async_session_factory() as session:
+            result = await session.execute(delete(PushSubscription).where(PushSubscription.username == username))
             await session.commit()
             return result.rowcount
 
--- a/src/web/main.py
+++ b/src/web/main.py
@@ -14,7 +14,7 @@ import os
 import secrets
 import time
 import traceback
-from collections.abc import AsyncGenerator
+from collections.abc import AsyncGenerator, Iterable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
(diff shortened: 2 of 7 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #311, opened 2026-08-16
docs: merging two existing archives into one multi-account install

GeiserX:
## The problem

8.0 lets one archive hold many accounts, but the upgrade guide only covers adding a **fresh** second account — which starts from zero: a new login re-captures what Telegram still has, silently loses every message deleted since, every edit version, and re-downloads years of media. The other real story — two people each ran their own single-account install for years and now want ONE archive — had no path at all.

## What this adds

**docs/MERGING-ARCHIVES.md** + executable companion SQL under **scripts/merge/**, linked from the upgrade guide's multi-account section:

- **PostgreSQL→PostgreSQL and SQLite→SQLite fully**, mixed backends routed through the existing SQLite→PostgreSQL mover (verified 8.0-schema-aware end to end).
- The procedure is **auditable SQL the user can read** before running it on the only copy of their history — not a tool. One transaction per backend; explicit-column INSERT..SELECT remapped into a freshly assigned account row; per-table expected-vs-actual recounts; FK re-verification.
- **Every unsafe state aborts loudly with a named fix**: mismatched schema revisions, an install never started (unclaimed account row), two archives of the same Telegram account (that's message-level dedupe — refused honestly), a chat ref collision, another connection holding the database.
- **SQLite deliberately does not commit**: the script ends with a verdict line and the last word is the human's — `COMMIT;` on OK, `ROLLBACK;` otherwise, and quitting rolls back too. PostgreSQL commits only if every DO-block check passed.
- What merging preserves over a fresh login is stated up front: deleted-message history, edit versions, already-downloaded media, and sync cursors — so the merged account's first sweep is incremental instead of a years-long re-capture.
- Media transfer (`rsync -a --ignore-existing`, shared-store symlinks, path-prefix remap), the session-file **move** (never copy — a session used from two processes is burned), env declaration, first-start log verification (counts only), and viewer re-grants are all in the doc.

## Evidence

Two lanes, both green. The author proved the procedure live on synthetic two-account archives on both backends plus the mixed path — including the shared-group case (same chat, overlapping message ids, identical edit hashes in both archives: 022's motivating collision) — with target-side rows byte-stable under per-table digests and `PRAGMA foreign_key_check` / anti-join FK checks clean. An independent reviewer then **executed the committed doc verbatim** on their own fixtures, both backends, all green. Every guard was watched go red: three sabotage runs per backend (wrong revision, same-account, planted ref collision) each aborted with the target byte-identical, a missing-parameter control caught a silent-exit bug (fixed to a raising DO block), and a mutation control with a crippled script proved the count verification refuses commit. All script output is counts-only — no chat ids, titles, or content.

## Summary by CodeRabbit

- **New Features**
  - Added tools for safely merging a single-account archive into an existing multi-account archive.
  - Supports SQLite and PostgreSQL databases with validation, integrity checks, account remapping, and rollback protection.
  - Preserves historical data, media references, users, and chat relationships during the merge.
  - Provides clear commit-or-rollback verification for SQLite and automatic transaction rollback for PostgreSQL.

- **Documentation**
  - Added comprehensive merge instructions, including prerequisites, backups, verification, recovery guidance, and permissions.
  - Updated the 8.0 upgrade guide to direct users with existing archives to the merge workflow.

### DIFF
--- a/docs/MERGING-ARCHIVES.md
+++ b/docs/MERGING-ARCHIVES.md
@@ -111,8 +111,9 @@ cleanly stopped in step 2, so the file is complete without sidecar files; if
 `-wal`/`-shm` files exist anyway, copy them alongside).
 
 Run the script **interactively** — it deliberately does not commit. You need
-the `sqlite3` command-line shell, version 3.16 or newer (`apt install sqlite3`
-on Debian/Ubuntu; macOS ships one):
+the `sqlite3` command-line shell, version 3.33.0 or newer — the script uses
+`.mode box` for its report (`apt install sqlite3` on Debian/Ubuntu; macOS
+ships one):
 
 ```bash
 cd telegram-archive          # the checkout containing scripts/merge/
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
All three addressed in the follow-up commit: the SQLite floor is now 3.33.0 (`.mode box` — good catch, the earlier 3.16 figure was wrong), the psql example reads the password with `read -rs` and the doc states the remaining argv exposure honestly (with the passwordless scratch-database route recommended for shared machines), and the media-root probe is non-greedy — verified on PostgreSQL 18 that `'^(.*?/media/)'` captures to the first `/media/` exactly like the SQLite script's `instr()`: a two-`/media/` path now yields the same verdict on both backends.

PR #316, opened 2026-08-19
fix(db): drop redundant idx_messages_chat_id index

jordanfelle:
## Summary

- `idx_messages_chat_id` (plain `btree(chat_id)`) is a strict prefix of two later composite indexes — `idx_messages_chat_id_id (chat_id, id)` added in 017 and `idx_messages_chat_date_desc (chat_id, date DESC)` added in 005. Every `Message.chat_id ==` filter site in `src/db/adapter.py` also orders or bounds by `id`/`date`/`is_pinned`, so those composites already serve the plan; nothing needs the standalone index.
- Confirmed live on a ~385k-row archive via `pg_stat_user_indexes`: `idx_messages_chat_id` sat at 0 scans while the two composites carried hundreds of thousands of scans between them. It also carries none of the "why this exists" comments every neighboring index in `models.py` has, consistent with it being unmaintained cruft from the initial schema (001) rather than an intentional addition.

## Type of Change

- [ ] Documentation update

## Database Changes

- [x] Schema changes (Alembic migration required) — adds `025`, drops the index; `downgrade()` recreates it for safety.
- [ ] No database changes

## Data Consistency Checklist

- [ ] All `chat_id` values use marked format (via `_get_marked_id()`) — n/a, no `chat_id` values are read or written by this change
- [ ] All datetime values pass through `_strip_tz()` before DB operations — n/a, no datetime values touched
- [ ] INSERT and UPDATE operations handle the same fields identically — n/a, no INSERT/UPDATE paths touched

## Testing

- [x] Tests pass locally (`python -m pytest tests/ -v`) — 3244 passed, 109 skipped, 0 failed
- [ ] Manually tested in development environment — verified against a live ~385k-row PostgreSQL archive (index usage stats, `DROP INDEX` applied and confirmed), not a fresh dev environment

`tests/test_migration_025.py` covers: drops the index, is idempotent on a database that never had it, downgrade restores it (and the recreated index has the correct `column_names`), survives a missing `messages` table, and asserts the ORM model no longer declares it. `tests/test_schema_parity.py` (ORM vs. Alembic-built schema) still passes.

## Security Checklist

- [x] User input properly validated/sanitized — n/a, no user input in this change
- [x] Authentication/authorization properly checked — n/a, no auth paths touched

## Deployment Notes

Standard `alembic upgrade head` on next deploy. No Docker image rebuild needed beyond picking up the new migration file. `downgrade()` is safe to run if ever needed — it just recreates the dropped index.

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed a redundant message index to streamline database indexing while preserving composite chat-based indexes.
  - Ensured database migrations safely handle repeated upgrades, downgrades, and schemas without a messages table.

- **Tests**
  - Added coverage for index removal, preservation of unrelated indexes, migration sequencing, idempotency, downgrade recovery, and missing-table scenarios.

### DIFF
--- a/alembic/versions/20260819_025_drop_redundant_chat_id_index.py
+++ b/alembic/versions/20260819_025_drop_redundant_chat_id_index.py
@@ -1,0 +1,59 @@
+"""Drop idx_messages_chat_id - a strict prefix of two later composite indexes.
+
+``idx_messages_chat_id`` (chat_id) has existed since 001 and, unlike every
+other index on this table, carries no comment explaining what it serves.
+Every access path that filters ``Message.chat_id`` on its own also orders or
+bounds the result (by date, by id, or by is_pinned) and so already seeks
+through ``idx_messages_chat_id_id`` (chat_id, id, added 017) or
+``idx_messages_chat_date_desc`` (chat_id, date DESC, added 005) - both are a
+superset of this index's leading column, so the planner can use either for a
+bare chat_id predicate too. Checked every ``Message.chat_id ==`` /
+``Message.chat_id.__eq__`` site in src/db/adapter.py (get_messages_paginated,
+get_pinned_messages, toggle_pinned, delete_chat_messages, the per-day
+export branches): none of them stop at chat_id alone without also ordering
+or bounding by date/id/is_pinned.
+
+Confirmed via ``pg_stat_user_indexes`` on a live ~385k-row archive:
+``idx_messages_chat_id`` sat at 0 scans while ``idx_messages_chat_id_id`` and
+``idx_messages_chat_date_desc`` carried hundreds of thousands between them -
+the planner was already skipping the narrower index in favor of the
+composites. Dropping it removes 3.2MB of on-disk index and one write on
+every message insert, with no query plan depending on it.
+
+Named explicitly rather than reflecting models.py, matching 024's rationale:
+a migration has to keep meaning what it meant when it was written.
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "025"
+down_revision: str | None = "024"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+TABLE_NAME = "messages"
+INDEX_NAME = "idx_messages_chat_id"
+
+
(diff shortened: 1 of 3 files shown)


GeiserX (commented):
Verified end to end against this repo and against a second live archive — this holds up, and it's a model contribution.

What I checked beyond CI:

- **The redundancy claim.** `idx_messages_chat_id` (models.py:181) is a strict prefix of `idx_messages_chat_id_id` (:188) and `idx_messages_chat_date_desc` (:185), and a repo-wide grep finds no live reference to it — only the historical creates in 001/005, which correctly stay untouched. The FK side stays covered too: `fk_messages_chat` is `(account_id, chat_id)`, and the composite primary key serves those lookups.
- **Stamping.** `scripts/entrypoint.sh` deliberately stamps pre-Alembic schemas at 018 and lets every later guarded migration run idempotently — your existence-guarded drop slots into that with no stamping change needed, including the `create_all()` path where the index never existed. `test_a_database_already_missing_it_is_untouched` covering the 024-drift case was a nice touch.
- **Your `pg_stat` evidence, replicated on a larger archive.** On a 1.78M-row dual-account production instance the cumulative counter initially looked like a refutation — ~226k scans on this index. It turned out the entire number dates from a two-day window in which `idx_messages_chat_date_desc` and `idx_messages_chat_pinned` were missing (exactly the drift migration 024 now heals) and the narrow index was the only chat_id-leading option the planner had. Since the declared indexes were restored, a measured hour of live viewer traffic put **+5,558 scans on `chat_date_desc`, +796 on `chat_pinned`, +136 on `chat_id_id` — and exactly 0 on `idx_messages_chat_id`**. Steady state matches your zero.

One thing worth saying out loud: during that drift window the narrow index was acting as an accidental safety net. With 024 restoring the declared indexes automatically, that net is redundant — which is precisely your argument.

Thanks for reading the codebase this carefully. The migration docstring in the house style, the honest test for the drift case, and evidence pulled from a real archive made this a pleasure to verify.

GeiserX:
Released in [v8.0.3](https://github.com/GeiserX/Telegram-Archive/releases/tag/v8.0.3) and deployed — migration 025 ran clean on the production archive from the review (upgrade 024 → 025, index gone, zero errors). Thanks again.

PR #332, opened 2026-08-20
fix(backup): stream media verification instead of materialising the whole table

GeiserX:
## Problem

`VERIFY_MEDIA` loaded **every media row of the account into memory at once** — full ORM instances plus a dict per row — before checking a single file. The adapter itself records the consequence: this exact materialisation OOM-killed the 256m backup container on a large archive. Repair (`iter_media_paths_for_repair`) was switched to keyset streaming for precisely that reason; verification kept the old path.

## Fix

`get_media_for_verification` is replaced by `iter_media_for_verification`: keyset pagination on the string media id (unique within one account — same proven pattern as the repair iterator), projecting only the eight columns verification consumes, yielded in 500-row batches. Phase 1 consumes the stream and keeps only records that actually need re-downloading, so resident memory is bounded by batch size plus actual damage, not archive size. Phase 2 (re-download) is untouched — it already groups the issue list by chat itself, so the changed row order is irrelevant.

The "Found N media records to verify" upfront log becomes "Checked N media records to verify" after the pass — the count isn't known until the stream ends.

## Verification

- New `tests/test_verify_stream.py` on real engines: exact-once keyset coverage across batch boundaries (`[2,2,1]` batches, ordered, no dups), column projection, an exact batch-size multiple (clean tail probe), account scoping.
- Existing verification behavior tests (missing / corrupted / symlink / poison / account isolation) all pass through a per-file bridge that yields their flat-list seeding as one batch — production consumes only the iterator.
- Mutation controls: advancing the keyset from the wrong end (duplicates) and dropping missing-file collection each turn tests red; files restored byte-identical.
- Full suite: 3351 passed; `ruff check` + `ruff format --check` (0.15.14) clean.

## Summary by CodeRabbit

- **Performance**
  - Media verification now processes records in manageable batches, reducing memory usage and improving reliability for large media libraries.

- **Bug Fixes**
  - Preserved detection of missing files, broken symlinks, empty files, and file-size mismatches.
  - Verification remains correctly scoped to the selected account and processes each record exactly once.

- **Tests**
  - Added coverage for batching, ordering, account isolation, and large verification runs.

### DIFF
--- a/src/db/adapter.py
+++ b/src/db/adapter.py
@@ -2022,34 +2022,60 @@ class DatabaseAdapter:
             await session.commit()
             return result.rowcount
 
-    async def get_media_for_verification(self, *, account_id: int) -> list[dict[str, Any]]:
-        """
-        Get one account's media records that should have files on disk.
-        Used by VERIFY_MEDIA to check for missing/corrupted files — the caller
-        re-downloads what is missing, and only this account's session can.
+    async def iter_media_for_verification(self, *, account_id: int, batch_size: int = 500):
+        """Yield batches of one account's media records that should have files
+        on disk (``downloaded=1`` OR ``file_path`` set). Used by VERIFY_MEDIA —
+        the caller re-downloads what is missing, and only this account's
+        session can.
 
-        Returns media where downloaded=1 OR file_path is not null.
+        Keyset-paginated on ``id`` (a string, unique within one account),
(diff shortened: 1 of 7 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

Issue #95, opened 2026-03-25
[Bug]: Stuck at Fetching dialog list


sube32:
### Version

v7.3.0

### Deployment Method

Docker Compose

### Database Type

SQLite (default)

### Bug Description

Telegram-backup cannot fetch dialog list, gets stuck, no feedback.

### Expected Behavior

Any feedback.

### Steps to Reproduce

_**sudo docker compose up -d && sudo docker compose logs -f telegram-backup**_

```
telegram-backup  | No database found yet - skipping migrations (will be created automatically)
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - ============================================================
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - Telegram Backup Automation
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - ============================================================
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - Schedule: 0 */6 * * *
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - Backup path: /data/backups
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - Download media: True
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - Chat types: private, groups, channels
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - Real-time listener: disabled
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - ============================================================
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - Establishing shared Telegram connection...
telegram-backup  | 2026-03-25 11:03:09 - src.connection - INFO - Connecting to Telegram...
telegram-backup  | 2026-03-25 11:03:09 - src.connection - INFO - Enabled WAL mode for Telethon session database
telegram-backup  | 2026-03-25 11:03:09 - src.connection - INFO - Connected as *** (***)
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - Shared connection established
telegram-backup  | 2026-03-25 11:03:09 - apscheduler.scheduler - INFO - Adding job tentatively -- it will be properly scheduled when the scheduler starts
telegram-backup  | 2026-03-25 11:03:09 - src.scheduler - INFO - Backup scheduled with cron: 0 */6 * * *
[...]

PR #334, opened 2026-08-20
fix(backup): stop whitelist mode wiping every chat's archived flag with a fabricated 0

GeiserX:
## Problem

Whitelist mode skips `get_dialogs` on purpose (#95: the full fetch can hang on large accounts) and declared empty placeholders (`archived_chat_ids = set()`) so the shared downstream code would run. The side effect: `is_archived` computed `False` for **every** whitelisted chat on **every** run and was written unconditionally. Consequences:

- the viewer's Archived section is permanently empty in whitelist mode (`/api/archived/count` returns 0);
- switching from type-based mode to `CHAT_IDS` destroys the correctly-computed flags on the next run;
- the adapter guard that exists precisely to protect `is_archived` from partial writers was defeated by the key simply being present.

## Fix

- **Real membership, no dialog listing:** one batched `GetPeerDialogs` pass over exactly the resolved peers (100 per request, flood-wrapped) — bounded by whitelist size, not account dialog count, so the #95 constraint holds. `folder_id == 1` → archived.
- **Tri-state `is_archived`:** `True`/`False` write `1`/`0` exactly as before; `None` (probe failed) omits the key entirely, so `upsert_chat`'s presence guard preserves stored values — never a fabricated 0. New chats take the schema default (0) on the insert branch.
- Type-based mode's computation (including the both-folders warning) is untouched.

## Verification

- New `tests/test_whitelist_archived.py` (7 tests): folder-1-only marking with realistic marked ids, 100-per-request batching (101 peers → 2 calls), probe failure → `None` (never an empty set), tri-state extraction (None omits the key), and `backup_all` end-to-end in whitelist mode feeding `True` / `None` into `_backup_dialog`.
- Mutation controls: counting any folder as archived, unconditional key write, and hardcoding `False` in the loop each turn tests red; file restored byte-identical.
- Existing whitelist (#95), chat-migration and `_extract_chat_data` suites: 174 passed. Full suite + ruff 0.15.14 clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved whitelist backup handling for archived chats and folders.
  * Preserved existing archived status when archived-folder checks cannot be completed.
  * Prevented unavailable archived status from being incorrectly recorded as unarchived.
* **Tests**
  * Added coverage for folder filtering, large peer sets, lookup failures, and archived-status handling.

### DIFF
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -1352,6 +1352,16 @@ class TelegramBackup:
                     has_synced_before = True
                     break
 
+            # Whitelist mode resolves entities without any dialog listing
+            # (#95: the full fetch can hang), so archived-folder membership
+            # comes from one batched GetPeerDialogs pass over exactly the
+            # resolved peers. None means the probe failed: no is_archived is
+            # written at all, preserving stored values instead of forcing a
+            # wrong 0 on every run.
+            archived_membership: set[int] | None = None
+            if self.config.whitelist_mode and filtered_dialogs:
+                archived_membership = await self._fetch_archived_membership([d.entity for d in filtered_dialogs])
+
             # Backup each dialog
             # v6.2.0: Check archived_chat_ids so chats in both INCLUDE_CHAT_IDS
             # and the archived folder get the correct is_archived flag immediately.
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

PR #338, opened 2026-08-20
fix(viewer): ship only the dependencies the viewer can actually import

GeiserX:
## Problem

The viewer image ran the same `uv sync` as the backup image, so the **internet-facing** container carried `telethon`, `cryptg`, `python-socks`, `APScheduler` and `psycopg2-binary` — none of which any module it ships can import. Verified against the tree: no `psycopg2`/`alembic`/`telethon` import exists anywhere under `src/web`, `src/db`, `config.py`, `realtime.py`, or the healthcheck (the image doesn't even copy `alembic/`, and PostgreSQL goes through `asyncpg`). An exposed image should carry only the code paths it can execute — every extra package is attack surface and pull weight.

## Fix

- `pyproject.toml` gains a PEP 735 dependency group, **`viewer-runtime`** (13 entries), and `Dockerfile.viewer` installs **only** that group: `uv sync --locked --no-install-project --only-group viewer-runtime`.
- Every group entry mirrors the exact requirement string in `[project] dependencies`; `tests/test_viewer_dependency_group.py` enforces the subset verbatim (drift = red) and bans the capture stack from the group by name.
- The redundant `[viewer]` extra (Pillow only — already a base dependency — referenced nowhere in CI, Dockerfiles, or docs) is removed.
- `uv.lock` regenerated: purely additive diff, lock revision unchanged (still `3`, readable by the pinned `uv:0.11` in both Dockerfiles).

## Verification (real environments, not inference)

- A venv synced with `--only-group viewer-runtime` **imports `src.web.main`, `src.web.thumbnails`, `src.realtime`, `src.db.adapter` successfully** (with `BACKUP_PATH` writable), including the WebSocket/push/thumbnail paths (`websockets`, `pywebpush`/`py-vapid`/`cryptography`, `Pillow`).
- Control: removing Pillow from that venv makes the import fail — the check can go red.
- This PR's own `build-and-push-dev` check builds `Dockerfile.viewer` with the new sync line — the CI-side proof.
- Full suite green locally (3351 passed, gated with pipefail); ruff 0.15.14 check+format clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved viewer image builds by installing only the dependencies required at runtime.
  * Excluded capture-related packages from the viewer runtime image, reducing unnecessary dependencies.

* **Chores**
  * Updated publishing workflows to rebuild when viewer dependency definitions change.
  * Replaced the legacy viewer requirements file with a locked runtime dependency group.

* **Tests**
  * Added validation ensuring viewer runtime dependencies remain aligned with project requirements and exclude capture packages.


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #343, opened 2026-08-20
fix(compose): real-time pushes reach the viewer container on the stock SQLite stack

GeiserX:
## Problem

With `ENABLE_LISTENER=true` on the shipped compose (default SQLite backend), the backup container posts realtime events to `http://localhost:8080/internal/push` — **its own loopback** — while the viewer is a separate container listening on `0.0.0.0:8000`. Every push is refused and swallowed as a warning, so the README's headline *"Real-time updates — WebSocket sync shows new messages instantly"* silently degrades to refresh-to-see. The escape hatch documented in `.env.example` (`VIEWER_HOST=telegram-viewer` / `VIEWER_PORT=8000`) could not work either: compose never forwarded those variables into the container (also fixed generally by #333's `env_file`; this PR makes the CORRECT values the default so nobody needs to know the knob exists).

## Fix

- `telegram-backup` now carries `VIEWER_HOST: ${VIEWER_HOST:-telegram-viewer}` and `VIEWER_PORT: ${VIEWER_PORT:-8000}` — pushes target the viewer service by name on the compose network, overridable as before.
- README's `VIEWER_HOST`/`VIEWER_PORT` rows now state the `localhost` defaults only fit a same-host (bare-metal) setup and that the shipped compose overrides them.

## Verification

- Compose-wiring guard test asserts both lines; removing them turns it red (mutation-checked). yamllint clean.
- Note for the merge train: touches `docker-compose.yml` and `test_graceful_shutdown.py` like #333 — whichever merges second gets a trivial rebase.

## Summary by CodeRabbit

* **New Features**
  * Enabled real-time push delivery between the backup and viewer services in Docker Compose deployments.

* **Documentation**
  * Updated environment-variable documentation with the viewer service defaults.

* **Tests**
  * Added coverage to verify the services are correctly connected for real-time updates.


GeiserX:
@coderabbitai full review

PR #346, opened 2026-08-21
fix(config): one boolean vocabulary for every flag, and DB_TYPE case can no longer skip migrations

GeiserX:
Two config footguns that fail silently in opposite directions.

Thirteen boolean flags accepted only the literal string "true": DOWNLOAD_MEDIA=1 turned all media capture off, ENABLE_LISTENER=yes turned real-time capture off — no warning, nothing in the logs, while four sibling flags happily accepted 1/yes/on. And in the mirror direction, a typo on those four raised "Invalid boolean value: ture" without naming which variable carried the typo.

Every flag now parses one vocabulary (1/true/yes/on — 0/false/no/off) through a single parser, and a bad value names its variable in the error. scripts/entrypoint.sh additionally lower-cases DB_TYPE before matching, so DB_TYPE=PostgreSQL no longer falls through the exact-match into the no-migration path and boots the app on an unmigrated schema — an unrecognised DB config now refuses to start loudly instead.

Regression tests cover the full vocabulary on representative flags, the named-variable error, and the entrypoint case normalisation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Database type settings are now handled consistently regardless of capitalization.
  * Unsupported or unrecognized database configurations now stop startup with a clear error instead of skipping migrations.
  * Boolean configuration values are validated consistently across all settings.
  * Invalid values now identify the affected configuration variable, making setup errors easier to diagnose.
  * Standard true/false values such as `yes`, `no`, `on`, and `off` are supported consistently.

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -e
 
+# DB_TYPE is lowercased everywhere in Python (src/db/base.py, alembic/env.py)
+# but compared literally below — DB_TYPE=PostgreSQL used to match NEITHER
+# branch, silently skipping migrations and starting the app against a
+# zero-table database. Normalise once so both worlds agree.
+DB_TYPE=$(printf '%s' "${DB_TYPE:-}" | tr '[:upper:]' '[:lower:]')
+
 # Determine if we should run migrations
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai full review

PR #350, opened 2026-08-21
perf(backup): video thumbnails warm at capture time instead of first view

GeiserX:
The gallery emits /media/thumb/ URLs for videos, but the backup-side pre-generator returned early for any non-image — so every video thumbnail was extracted on the request path, at first view: ~600ms per healthy clip, and a cold grid of 24 videos took ~7 seconds to fill behind the 2-slot ffmpeg lane. (The other halves of this finding already shipped separately: failures are cached with a 5-minute TTL, and concurrent duplicates collapse to one generation.)

Downloaded videos now pre-generate through the exact _generate_video_sync the viewer uses — its ffmpeg-availability check, 4x byte gate, -max_pixels decoder cap and atomic temp+replace save all apply — inside the backup's existing to_thread lane, so the event loop never waits on ffmpeg and the common case is warm before any viewer asks. The request path remains the fallback for anything captured before this change. The backup image already ships ffmpeg.

Deliberately NOT taken from the finding: dropping the second ffmpeg seek attempt when the first fails. Distinguishing 'undecodable' from 'shorter than the 1s seek' by exit code alone is not reliable across containers, and guessing wrong permanently costs a short video its thumbnail to save 15 seconds once per bad file per 5-minute failure-cache window.

Tests: routing to the video generator with exact (source, dest, size) args; the cached-thumbnail skip; a real end-to-end run building a 2s clip with ffmpeg and asserting a valid WebP (skipped where ffmpeg is absent — CI has it). The image-only mutant fails both routing tests. Full suite: 3354 passed, 127 skipped, 451 warnings, 691 subtests passed in 90.68s (0:01:30)

## Summary by CodeRabbit

* **New Features**
  * Added automatic thumbnail generation for supported video files.
  * Video thumbnails are generated in WebP format and reuse existing cached thumbnails when available.
  * Large video files are handled with appropriate size safeguards.

* **Bug Fixes**
  * Improved thumbnail pre-generation so videos are no longer skipped when processing media backups.

### DIFF
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -185,7 +185,9 @@ def _pre_generate_thumbnail(source_path: str, media_root: str) -> None:
         from src.web.thumbnails import (
             _IMAGE_EXTENSIONS,
             _MAX_SOURCE_BYTES,
+            _VIDEO_EXTENSIONS,
             WEBP_QUALITY,
+            _generate_video_sync,
             _thumb_path,
         )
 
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

PR #353, opened 2026-08-21
fix(viewer): a broken media volume no longer reports "Database temporarily unavailable"

GeiserX:
_is_db_connection_error exists to split "the database is unreachable" (retryable 503) from everything else (500) — but its first branch matched bare OSError, the base class of every filesystem, permission and disk-full error in Python. A read-only media volume, a full disk, or a thumbnail-cache mkdir failure all answered 503 {"detail": "Database temporarily unavailable"}: the operator gets sent to debug Postgres, and uptime monitoring or client retry logic keyed on 503 points at the wrong subsystem. Reproduced in the original finding with a NotADirectoryError from the thumbnail cache.

The predicate now matches only connection-shaped errors — ConnectionError and its subclasses plus TimeoutError, which is what a dead DB socket actually raises. Real DB connect and DNS failures on database paths were never relying on the OSError branch: SQLAlchemy wraps them in OperationalError, which the next branch already catches (asyncpg's own connection errors have their dedicated branch too).

Filesystem faults now take the 500 branch. That branch redacts exactly like the 503 one — describe_exception refuses path-carrying exception text and only code frames are logged — so the move leaks nothing; the ACL redaction suite now pins the no-identifier property on the 500 branch for a FileNotFoundError carrying a full media path, and keeps pinning the 503 branch with a genuine connection error. Tests across the suite that simulated a dead database with a bare OSError stand-in now use ConnectionRefusedError, keeping their 503 intent with an honest fixture.

Mutation control: restoring the OSError branch fails the filesystem-fault tests. Full suite 3365 passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database outage detection so filesystem and disk-related errors are no longer incorrectly treated as connectivity failures.
  - Preserved HTTP 503 responses for genuine database connection refusals, timeouts, resets, and related errors.
  - Filesystem “not found” errors now correctly return HTTP 500 responses.

- **Tests**
  - Expanded coverage for direct and chained database connection errors across web routes and exception handling.

### DIFF
--- a/src/web/main.py
+++ b/src/web/main.py
@@ -879,7 +879,13 @@ def _is_db_connection_error(exc: Exception) -> bool:
     for _ in range(10):
         if current is None:
             break
-        if isinstance(current, OSError):
+        # Only connection-shaped OS errors count. Bare OSError is the base of
+        # every filesystem fault (NotADirectoryError, PermissionError, ENOSPC),
+        # and matching it here sent operators to debug the database while the
+        # media volume or thumbnail cache was the broken part. DB connect and
+        # DNS failures on database paths arrive wrapped in OperationalError,
+        # which the next branch already catches.
+        if isinstance(current, ConnectionError | TimeoutError):
             return True
         if isinstance(current, OperationalError):
             return True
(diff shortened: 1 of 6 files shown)


GeiserX:
@coderabbitai full review

PR #354, opened 2026-08-21
fix(backup): MAX_MEDIA_SIZE_MB finally caps full-resolution photos

GeiserX:
An operator capping disk usage with MAX_MEDIA_SIZE_MB gets no cap on photos at all. Telegram delivers a big photo's full-resolution rendition as PhotoSizeProgressive, which has no scalar .size — only a sizes list of progressive byte offsets. The backup path read getattr(sizes[-1], 'size', 0) → 0, and the listener picked the largest rendition by getattr(s, 'size', 0) → the progressive entry scores 0 and a small thumbnail wins. Either way a 9.5 MB photo measures as 0 bytes (backup) or ~90 KB (listener) and sails through a 5 MB cap. Secondary effect: _should_parallelize receives file_size=0, so parallel download never engages for large photos.

The repository already contains the correct answer: message_utils._photo_size_bytes, whose docstring documents exactly this progressive shape (it is what fixed the same scoring bug for dimension metadata). Both ingest paths now use it, taking the max across renditions instead of trusting list position. Three implementations of one question became one.

Tests build the real Telethon types (PhotoSize + PhotoSizeProgressive, the finding's exact 9.5 MB scenario) on both paths, including a progressive-first ordering and a listener gate that must refuse the download outright. Both old implementations, restored as mutants, fail the new tests. Full suite: 3366 passed, 127 skipped, 447 warnings, 691 subtests passed in 88.47s (0:01:28)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved photo size detection across all available renditions, including progressive photos.
  * Media-size limits now correctly prevent downloads when the full-resolution photo exceeds the configured cap.

* **Tests**
  * Added coverage for progressive photo sizes and varying rendition order.

### DIFF
--- a/src/listener.py
+++ b/src/listener.py
@@ -39,6 +39,7 @@ from .config import AccountConfig, Config
 from .db import DatabaseAdapter, create_adapter
 from .db.models import account_metadata_key
 from .message_utils import (
+    _photo_size_bytes,
     build_media_filename,
     compute_file_hash_async,
     describe_exception,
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai full review

PR #356, opened 2026-08-21
fix(backup): excluding a chat purges it from the archive no matter which folder it lives in

GeiserX:
backup_all documents that putting a chat in an exclude list deletes it and its media from the archive, and the deletion loop does exactly that — for chats in the regular dialog list. The excluded set was built only there: a chat the user had archived in Telegram (folder 1) never entered it, so the archived loop just skipped the chat and moved on. The identical user action therefore meant two different things depending on which Telegram folder the chat sat in: full purge for a regular chat, silent keep-forever for an archived one, with every message row, reaction, media row and media file staying on disk — the storage the user was trying to reclaim.

The explicit-exclusion predicate now lives in one helper (_is_explicitly_excluded) applied to both dialog lists before the deletion loop runs, so the archived half is purged on the same run and can never drift from the regular half again. Excluded ids were already subtracted from the missing-include fetch, so that stays coherent.

Tests: an archived-only chat in global_exclude_ids is deleted (chat id asserted); a control pins that a merely filtered-out archived chat is never deleted. The no-archived-pass mutant fails the first test — verified with the unmutated selector proven green first. Full suite 3365 passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Explicitly excluded chats are now consistently removed, including archived groups, channels, bots, and users.
  - Archived chats that are not excluded remain preserved during backups.
- **Tests**
  - Added coverage for handling excluded and non-excluded archived chats.

### DIFF
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -978,6 +978,24 @@ class TelegramBackup:
             await self.client.disconnect()
             logger.info("Disconnected from Telegram")
 
+    def _is_explicitly_excluded(self, chat_id: int, entity) -> bool:
+        """True when the user put this chat in an exclude list (not merely filtered out).
+
+        THE single copy of the predicate: the regular and archived dialog
+        passes must agree, or exclusion means "purge" in one Telegram folder
+        and "keep forever" in the other.
+        """
+        is_bot = isinstance(entity, User) and entity.bot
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

PR #358, opened 2026-08-21
fix(listener): the realtime notifier rides the listener's own engine, not the process global

GeiserX:
TelegramListener.create() builds its adapter, then connect() — several awaits later (authorization check, chat-list load) — re-resolved the process-global db manager to hand the realtime notifier. Both long-lived components reassign that global: every scheduled backup run replaces it with a fresh engine and disposes it in its run-end close(). A backup starting inside connect()'s await window therefore bound the listener's notifier to an engine an unrelated component would tear down — on PostgreSQL that silently drops the pooled connections the notifier was using (SQLAlchemy lazily rebuilds after dispose, so it degrades to connection churn rather than hard failure).

The notifier now binds to self.db.db_manager — the exact manager the listener's own adapter already owns — so no other component's lifecycle can touch it, and the re-resolution window is gone entirely.

Deliberately NOT taken: the finding's larger remedy of moving both components off the global-mutating create_adapter() onto injected private managers. With each consumer now holding its own reference, what remains of the global-swap pattern costs reconnection churn, not correctness — a bootstrap refactor is not worth that residue. 

The regression test swaps the global mid-connect (a foreign manager behind a patched get_db_manager) and pins that the notifier still holds the listener's own manager and the global resolver is never consulted; the rebind-to-global mutant fails it (control proven green unmutated first). Full suite 3368 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved real-time notifications by ensuring they use the listener’s configured database connection.
  * Prevented notifications from incorrectly switching to a process-wide database connection, improving reliability in multi-connection environments.

* **Tests**
  * Added regression coverage to verify notifications remain bound to the listener’s own database manager and do not use a global connection.

### DIFF
--- a/src/listener.py
+++ b/src/listener.py
@@ -448,11 +448,12 @@ class TelegramListener:
         # Load tracked chat IDs from database
         await self._load_tracked_chats()
 
-        # Initialize real-time notifier (auto-detects PostgreSQL vs SQLite)
-        from .db import get_db_manager
-
-        db_manager_instance = await get_db_manager()
-        self._notifier = RealtimeNotifier(db_manager_instance)
+        # Initialize real-time notifier (auto-detects PostgreSQL vs SQLite).
+        # Bound to THIS listener's own manager — never re-resolved from the
+        # process global: a cron backup starting inside connect()'s await
+        # window reassigns that global with a fresh engine, and its run-end
+        # dispose() would then tear the pool down under the notifier.
+        self._notifier = RealtimeNotifier(self.db.db_manager)
         await self._notifier.init()
         logger.info("Real-time notifier initialized")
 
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

GeiserX:
Reviewed the nitpick: its premise doesn't hold here — neither `src/listener.py` nor `src/realtime.py` imports `get_db_manager` at all (the old bug was a no-arg `RealtimeNotifier()`, which left `_db_manager` None for env auto-detection, not a direct import). The load-bearing assertion is `listener._notifier._db_manager is own_manager`, which catches any future regression shape independent of patch site; the patched resolver is only a belt-and-braces canary. No change needed.

PR #362, opened 2026-08-21
fix(backup): quoted-reply excerpts are finally captured

GeiserX:
The sweep's reply-preview branch read reply_to.message — an attribute telethon's MessageReplyHeader has never had (verified against 1.43.2: its text field is quote_text) — so the hasattr guard could never fire and the sweep stored reply_to_text as None for every reply. The branch read as working code, and its only covering test kept it green by building the impossible shape with a bare MagicMock — the exact truthiness pitfall this repo's CLAUDE.md documents.

For plain replies the damage is masked: the viewer backfills the preview at read time from the local reply row, failing only when the target isn't archived. But a QUOTE reply — where the sender quoted a fragment of the target, which official clients render as the excerpt — cannot be reconstructed from the target's full text, so that excerpt was lost outright, permanently.

The branch now stores quote_text truncated to Telegram's own 100-char preview length, exactly when the sender actually quoted. The test is rewritten against the real MessageReplyHeader: the quote case pins content and truncation, the plain-reply case pins None (backfill's domain). The restored dead-branch mutant fails the quote test (proven green unmutated first). Full suite 3368 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reply previews now correctly capture quoted text, preserving up to 100 characters.
  * Ordinary replies no longer display incorrect reply text when no quote is included.

### DIFF
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -2972,13 +2972,17 @@ class TelegramBackup:
         if hasattr(message, "post_author") and message.post_author:
             message_data["raw_data"]["post_author"] = message.post_author
 
-        # Get reply text if this is a reply
+        # Quote-reply excerpt. MessageReplyHeader has NO ``message`` attribute
+        # (the old hasattr guard could never fire, so reply_to_text was always
+        # None from the sweep): its text field is ``quote_text``, set exactly
+        # when the sender quoted part of the target. That excerpt is what the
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #365, opened 2026-08-21
fix(api): an offset-bearing cursor or expiry now means the instant it names

GeiserX:
## Problem

Two API inputs parsed ISO-8601 datetimes and then stripped the timezone without converting — relabelling an instant instead of translating it:

- `GET /api/chats/{id}/messages?before_date=2026-01-15T12:00:00+02:00` compared 12:00 against `Message.date` (naive UTC), when the caller named 10:00 UTC. Two hours re-delivered at every page boundary; with negative offsets, a silently skipped window. The bundled SPA is unaffected (it echoes back naive strings) — this bites third-party consumers of the documented HTTP API.
- `POST /api/admin/tokens` with `expires_at: 2026-09-01T12:00:00+02:00` stored 12:00 naive; enforcement compares against `utcnow_naive()` (adapter.py), so the share token outlived the admin's intent by two hours.

The same file already does this right in `get_message_by_date` and `get_message_dates` (`astimezone(UTC)` before stripping) — these were the two remaining relabel sites, found by sweeping `main.py` for `replace(tzinfo=None)`.

## Fix

Convert to UTC before dropping tzinfo at both sites. Z-suffixed and naive inputs are byte-for-byte unchanged.

## Evidence

- Full suite: 3418 passed, 127 skipped, 861 subtests, exit 0.
- Mutation controls (green first, then red): re-relabelling the cursor fails the offset test while the Z test rightly stays green (strip and convert coincide at UTC); dropping the expiry conversion fails its test. Restores sha-verified.

Closes bead Telegram-Archive-9t6.5.77.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected timezone conversion for message date cursors.
  - Corrected token expiration handling so timestamps consistently represent the intended UTC instant.

- **Tests**
  - Added coverage for timezone-offset message cursors.
  - Added coverage for token expiration timestamps with timezone offsets.

### DIFF
--- a/src/web/main.py
+++ b/src/web/main.py
@@ -2308,9 +2308,9 @@ async def get_messages(
     if before_date:
         try:
             parsed_before_date = datetime.fromisoformat(before_date.replace("Z", "+00:00"))
-            # Strip timezone for DB compatibility
+            # Message.date is naive UTC — convert the instant, never just relabel it
             if parsed_before_date.tzinfo:
-                parsed_before_date = parsed_before_date.replace(tzinfo=None)
+                parsed_before_date = parsed_before_date.astimezone(UTC).replace(tzinfo=None)
         except ValueError:
             raise HTTPException(status_code=400, detail="Invalid before_date format. Use ISO 8601.")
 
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

PR #366, opened 2026-08-21
fix(db): importing an old export no longer re-fetches the whole chat from Telegram

GeiserX:
## Problem

`sync_status.last_message_id` is a high-water mark: the backup reads it as `min_id` for the next incremental pass. But `update_sync_status`'s ON CONFLICT clause assigned `excluded.last_message_id` unconditionally, so `telegram_import --merge` of an older export (whose max id is below the live checkpoint) dragged the cursor down. The next backup pass then re-fetched everything above it from Telegram — wasted API calls and real FloodWait risk on a large chat. Official clients keep their update checkpoints monotonic for exactly this reason.

## Fix

The conflict clause clamps the cursor: `max(stored, excluded)` on SQLite, `greatest(stored, excluded)` on PostgreSQL (no portable spelling exists — branched alongside the existing `_is_sqlite` split). `last_sync_date` and the `message_count` accumulation are unchanged. The backup's own calls always move forward, so they see no difference.

## Evidence

- Regression test in `test_db_adapter_real_engine.py` runs on BOTH backends (the PG leg executes on CI): 500 → 400 keeps 500; 600 still advances; count still accumulates.
- Mutation control (green first, then red): reverting the SQLite clamp fails the new test; restore sha-verified. The PG spelling is compile-verified to render `greatest(sync_status.last_message_id, excluded.last_message_id)`.
- Full suite: 3416 passed, 128 skipped, 861 subtests, exit 0.

Closes bead Telegram-Archive-9t6.5.67.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sync progress now retains the highest message ID, preventing older imports from moving the sync position backward.
  * Newer messages continue advancing the sync position while message counts accumulate correctly.

* **Tests**
  * Added coverage validating sync behavior with a real database engine.

### DIFF
--- a/src/db/adapter.py
+++ b/src/db/adapter.py
@@ -2481,7 +2481,10 @@ class DatabaseAdapter:
                 stmt = stmt.on_conflict_do_update(
                     index_elements=["account_id", "chat_id"],
                     set_={
-                        "last_message_id": stmt.excluded.last_message_id,
+                        # High-water mark: the backup reads this as min_id for the
+                        # next incremental pass, so it must never move backwards
+                        # (an older export import supplies a smaller max id).
+                        "last_message_id": func.max(SyncStatus.last_message_id, stmt.excluded.last_message_id),
                         "last_sync_date": stmt.excluded.last_sync_date,
                         "message_count": SyncStatus.message_count + stmt.excluded.message_count,
                     },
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #373, opened 2026-08-22
fix(backup): sync can no longer overwrite one message's text with another's

GeiserX:
## Problem

`_sync_deletions_and_edits` paired its `get_messages(ids=batch)` response with `zip(batch_ids, remote_messages)`. Telethon's `_IDsIter` buffers **one entry per RETURNED message**, and its own source comments that Telegram "may decide to not send" invalid ids at all on the non-channel path (users and basic groups; channels pad with `MessageEmpty`). One omission shifts every later position: message A's archived text gets overwritten with message B's, and a `None` landing on a shifted slot hard-deletes a message that still exists on Telegram (DELETION_MODE defaults to hard — row, media and reactions). Entirely silent: the run reports ordinary sync counts, and B905 (zip-without-strict) is explicitly ignored for this file.

## Fix

Positions are trusted only when the response **provably lines up** — same length AND every non-None entry's id matches its slot (the length check alone would still trust a reordered response). Otherwise the batch is keyed by id: matched messages process edits and reactions normally, and unmatched ids are treated as unknown for this run — logged as a count-only warning and retried next pass. A destructive write never rides an ambiguous signal.

## Evidence

- Verified against the installed Telethon's `_IDsIter._load_next_chunk` — the per-returned-message buffering and the omission warning are exactly as described.
- New regressions: an omitted mid-batch id deletes NOTHING and the surviving edit lands on its own id (the old zip overwrote the wrong row); a full-length but reordered response pairs by id. Mutation control forcing the old positional zip fails both; restore sha-verified.
- Existing mocks now carry real ids (a MagicMock id would silently divert them down the id-keyed path).

Closes bead Telegram-Archive-9t6.5.41.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization when Telegram returns messages in a different order than requested.
  * Prevented incorrectly deleting or overwriting messages when remote responses omit or reorder items.
  * Ensured unknown or missing message IDs are excluded from deletion processing.

### DIFF
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -2395,9 +2395,36 @@ class TelegramBackup:
                 # Fetch current state from Telegram
                 remote_messages = await call_with_flood_retry(self.client.get_messages, entity, ids=batch_ids)
 
-                for msg_id, remote_msg in zip(batch_ids, remote_messages):
+                # Telethon buffers one entry per RETURNED message, and its own
+                # comment warns Telegram "may decide to not send" invalid ids at
+                # all on the non-channel path — one omission shifts every later
+                # position, and a positional zip would overwrite message A's
+                # text with message B's or hard-delete a live message. Trust
+                # positions only when the response provably lines up; otherwise
+                # key by id and treat unmatched ids as unknown — a destructive
+                # write never rides an ambiguous signal.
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #374, opened 2026-08-22
fix(listener): renaming a supergroup no longer files it as a broadcast channel

GeiserX:
## Problem

Any title or photo change makes the listener refresh the chat's metadata — and it classified the entity with `hasattr(entity, 'broadcast')`. Telethon's `Channel` type always CARRIES that attribute (it is `False` on a megagroup, not absent), so every supergroup was written back as `type='channel'`. Folder membership resolves from the stored type, so on the next folder sync the supergroup matched the folder's `broadcasts` flag instead of `groups` and moved into the wrong folders in the viewer. The next full sweep of the dialog repairs the row — but the damage is permanent for any chat the sweep no longer visits (left group, narrowed CHAT_IDS).

## Fix

The refresh calls the class's own `_get_chat_type` — the same vocabulary the scheduled sweep enforces (`'channel' if not entity.megagroup else 'group'`, Users → private, small chats → group).

## Evidence

- Regressions drive the real handler with REAL Telethon `Channel` entities: a megagroup title change stores `group`, a broadcast channel stores `channel`. Under the reverted `hasattr` classifier the supergroup test goes red (and the broadcast one rightly stays green). Restore sha-verified.

Closes bead Telegram-Archive-9t6.5.49.

## Summary by CodeRabbit

- **Bug Fixes**
  - Chat metadata now correctly distinguishes Telegram supergroups from broadcast channels.
  - Supergroups are classified as “group,” while broadcast channels are classified as “channel” after chat updates.

- **Tests**
  - Added coverage for metadata refreshes triggered by title changes.
  - Documented and verified the distinction between megagroup and broadcast chat types.

### DIFF
--- a/src/listener.py
+++ b/src/listener.py
@@ -1435,7 +1435,12 @@ class TelegramListener:
                         if entity:
                             chat_data = {
                                 "id": chat_id,
-                                "type": "channel" if hasattr(entity, "broadcast") else "group",
+                                # _get_chat_type, not hasattr(entity, "broadcast"):
+                                # Telethon's Channel always CARRIES broadcast (False
+                                # on a megagroup), so the hasattr test relabelled
+                                # every supergroup as a broadcast channel and folder
+                                # sync then filed it under the wrong folder flag.
+                                "type": self._get_chat_type(entity),
                                 "title": getattr(entity, "title", None),
                                 "username": getattr(entity, "username", None),
                             }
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #375, opened 2026-08-22
fix(backup): a forward from a channel finally stores an id the user can look up

GeiserX:
## Problem

`_extract_forward_from_id` returned raw `channel_id`/`chat_id` values, while every other persisted id in this project is the MARKED form (`user_id`, `-chat_id`, `-100<channel_id>`) — the convention this same file's migration pointers already follow via `get_peer_id`. A raw channel id lands in the user-id numeric space: the archive cannot tell 'forwarded from channel X' apart from 'forwarded from user Y' when ids collide, no query using the id the user actually knows ever matches, and the viewer renders the meaningless number verbatim. One row even mixed both conventions (`sender_id` is marked per Telethon's own docstring).

## Fix

The helper returns `get_peer_id(peer)` (TypeError → None for unknown peer shapes). Rows written before this fix keep their raw values — the positive spaces overlap, so no reliable remap exists; new captures are correct and old rows age out with re-sweeps.

## Evidence

- Tests repointed from the bug-pinning raw assertions to real `PeerUser`/`PeerChannel`/`PeerChat` objects asserting 12345 / -1000000099999 / -77777. Mutation control restoring the raw extraction fails both marked tests; restore sha-verified.
- Full suite: 3415 passed, 127 skipped, 861 subtests, exit 0.

Closes bead Telegram-Archive-9t6.5.50.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected forwarded message IDs for channels and group chats.
  * Prevented channel and chat IDs from conflicting with user IDs.
  * Improved handling when a forwarded peer’s ID cannot be determined.

### DIFF
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -2798,15 +2798,15 @@ class TelegramBackup:
 
         peer = message.fwd_from.from_id
 
-        # Handle different Peer types
-        if hasattr(peer, "user_id"):
-            return peer.user_id
-        if hasattr(peer, "channel_id"):
-            return peer.channel_id
-        if hasattr(peer, "chat_id"):
-            return peer.chat_id
-
-        return None
+        # Store the MARKED id (user_id, -chat_id, -100<channel_id>) — the
+        # convention every other persisted id in this project follows and the
+        # id the user can actually look up. Returning the raw channel/chat id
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #379, opened 2026-08-22
refactor(capture): two pieces of dead machinery that read as live policy

GeiserX:
## Problem

- `_newer_edit_date` (adapter.py) has zero callers, but sits at module level named exactly for the job the live edit rule does — with materially different semantics (strict `incoming > current`, refuses a null incoming date, no unchanged-text guard). The next person adding an edit entry point reaches for it and silently gets a different policy; the divergence stays invisible until edits drop or phantom edited markers appear.
- `backup_all` computed `has_synced_before` with one `get_last_message_id` SELECT **per dialog** on every run — worst on the very first full archive, where no chat short-circuits the loop — and nothing ever consumed the result. A coverage-only test class asserted the dead path, making it look live.

## Fix

Deletion only: the helper, the probe loop, and the coverage-only test class. No behavior replaced because none existed.

## Evidence

- Zero references remain (`grep -c` = 0 across src and tests for both names).
- Full suite: 3408 passed, 127 skipped, 861 subtests, exit 0 — nothing consumed either piece.

Closes beads Telegram-Archive-9t6.5.45 and Telegram-Archive-9t6.5.44.

## Summary by CodeRabbit

* **Refactor**
  * Removed unused internal synchronization checks from backup processing.
  * Simplified date-comparison logic without changing public functionality.

* **Tests**
  * Removed obsolete coverage for detecting previously completed backups.


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #382, opened 2026-08-22
refactor(web): push notifications stop hand-rolling the sender-name rule

GeiserX:
## Problem

`resolve_sender_display_name` is documented as the single source of truth for 'who sent this row' — so the message list, gallery, export and quote block can never drift apart — and the adapter uses it at three call sites. The Web Push path hand-rolled a FOURTH copy of the precedence chain, and paid a `get_user_by_id` DB round trip per notification for fields the listener already enriches into the WS payload (`first_name`/`last_name`/`username`, mirroring the API row shape). A future precedence change would silently miss the push path, and users would see a different sender in the notification than in the app.

## Fix

The push path feeds the shared helper from the payload first — the common path costs no DB query — with the lookup retained only for payloads that carry no user fields at all (and even that fallback now routes through the helper).

## Evidence

- Three regressions: an enriched payload resolves with ZERO db lookups; the capture-time snapshot still wins; a bare payload falls back to exactly one lookup. Mutation control blinding the helper to the payload fields fails the first; restore sha-verified.
- Full suite: 3418 passed, 127 skipped, 861 subtests, exit 0.

Closes bead Telegram-Archive-9t6.5.46.

## Summary by CodeRabbit

* **Bug Fixes**
  * Push notifications now display the sender’s name from available message details when provided.
  * Sender information captured with a message takes precedence, while database lookup is used only when necessary.
  * Notifications without sender details now handle the sender name gracefully.

* **Tests**
  * Added coverage for sender-name precedence and fallback behavior.

### DIFF
--- a/src/web/main.py
+++ b/src/web/main.py
@@ -35,7 +35,7 @@ from ..config import Config
 from ..db import DatabaseAdapter, close_database, get_db_manager, init_database
 from ..db.adapter import ChatScope, parse_entitlement_column
 from ..db.models import DEFAULT_ACCOUNT_ID, account_metadata_key
-from ..message_utils import describe_exception, media_display_filename
+from ..message_utils import describe_exception, media_display_filename, resolve_sender_display_name
 from ..realtime import RealtimeListener
 from .media_utils import THUMBNAIL_EXTENSIONS, legacy_folder_alternates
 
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #385, opened 2026-08-22
fix(backup): an attachment whose download yields no file is finally retried

GeiserX:
## Problem

Whether a failed media download was recoverable depended on **how** it failed. An exception inside `_process_media` persisted a `downloaded=0` row — retried by the pending drain every cycle. But a no-file outcome (`download_media` returned nothing, or `finalize_atomic_download` found no file) returned `None`: no media row at all, so the message was archived as though it never had an attachment. `_retry_pending_media_downloads` only drains rows, the capped-downloads counter never saw it, and once the sweep cursor passed the message, no path ever revisited it. The only signal was one generic warning.

## Fix

Both no-file branches (dedup and direct) return the same value-less `downloaded=0` row the exception path always has, via one shared `_failed_media_row` helper, with the warning now saying the failure was recorded for retry.

The listener's identical-looking shape is deliberately left alone: its messages sit ABOVE the backup cursor (`get_last_message_id` reads sync_status, which only backup runs advance), so the next scheduled sweep re-attempts their media — verified against the cursor source before scoping.

## Evidence

- Regressions on both branches: a no-file download leaves `{id, type, message_id, chat_id, downloaded: False}`; mutation control restoring the silent `return None` fails both. Restore sha-verified.

Closes bead Telegram-Archive-9t6.5.40.

## Summary by CodeRabbit

- **Bug Fixes**
  - Failed media downloads now remain visible as pending items for retry, including downloads that produce no file or encounter an error.
  - Improved consistency for media download failure handling across standard and deduplicated downloads.

- **Tests**
  - Added coverage confirming failed downloads remain retryable in both download paths.

### DIFF
--- a/src/telegram_backup.py
+++ b/src/telegram_backup.py
@@ -613,6 +613,20 @@ async def iter_messages_with_flood_retry(client, entity, *, min_id=0, **kwargs):
             await _reconnect_before_retry(client, "iter_messages")
 
 
+def _failed_media_row(media_id: str, media_type: str, message_id: int, chat_id: int) -> dict:
+    """The value-less media row a failed download leaves behind.
+
+    downloaded=0 is what makes the failure retryable: the pending-media drain
+    only sees rows, so a failure that leaves none is permanently silent."""
+    return {
+        "id": media_id,
+        "type": media_type,
+        "message_id": message_id,
+        "chat_id": chat_id,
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #387, opened 2026-08-22
fix(db): infinite scroll stops paying 1.7x round trips on media-heavy chats

GeiserX:
## Problem

`get_messages_paginated` applied `LIMIT` to the message×media outer join. A message can legitimately carry SEVERAL media rows — a JSON import writes `import_{chat}_{msg}` beside the live `{chat}_{msg}_{type}` row — so a page of 50 could deliver far fewer distinct messages: the bead's measured case, 10 three-media heads → 50 rows / 30 messages. The frontend dedupes by id, so nothing rendered twice; the cost was throughput (~1.7× the round trips to cover the same history, each re-running the whole page pipeline) and duplicated `raw_data` bytes serialized per extra row.

## Fix

Media leaves the page query entirely and arrives via ONE batched query keyed by the page's message ids — the exact shape the function already uses for version counts and reactions. One deterministic row attaches per message: downloaded beats pending, then lowest media id (the old join order was arbitrary). The User join stays (users.id is unique — it cannot multiply), account matching is carried out-of-band so the API row shape is unchanged, and `get_pinned_messages` was checked: it joins media the same way but its LIMIT applies to pinned flags (bounded, no measured shortfall) — left alone.

## Evidence

- Real-engine regression seeds the measured shape: limit=50 returns 50 DISTINCT messages newest-first (the join-restored mutant returns fewer — red); a second test pins the deterministic media pick (downloaded-preference and id tiebreak) and None for media-less messages.
- The query-hardening guard is repointed to pin BOTH directions: exactly one media read keyed by page ids, and a page statement that never touches media.
- Mutation control green-then-red; restore sha-verified. Full suite: 3417 passed, 129 skipped, 861 subtests, exit 0.

Closes bead Telegram-Archive-9t6.6.14.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed message pagination so pages consistently contain the requested number of distinct messages, even when messages have multiple media attachments.
  * Improved media selection by prioritizing downloaded media and applying deterministic fallback ordering.
  * Preserved account ordering when reversing forward-pagination results.
  * Ensured messages without media continue to display correctly.

### DIFF
--- a/src/db/adapter.py
+++ b/src/db/adapter.py
@@ -2979,31 +2979,22 @@ class DatabaseAdapter:
         """
         async with self.db_manager.async_session_factory() as session:
             # Build query with joins - v6.0.0: join on composite key
+            # No Media join here: a message can carry SEVERAL media rows (a
+            # JSON import writes import_{chat}_{msg} beside the live
+            # {chat}_{msg}_{type} row), and LIMIT applied to the multiplied
+            # join meant a page of 50 could deliver far fewer distinct
+            # messages — measured 50 rows / 30 messages with 10 three-media
+            # heads. Media is batch-attached below from the page's id set,
+            # the same shape versions and reactions already use. The User
+            # join stays: users.id is unique, so it cannot multiply.
             stmt = (
                 select(
                     Message,
                     User.first_name,
                     User.last_name,
                     User.username,
-                    Media.id.label("media_id"),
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai full review

GeiserX:
@coderabbitai full review

PR #393, opened 2026-08-22
test(db): the version-dedup ON CONFLICT clause can no longer be deleted with the suite green

GeiserX:
`_record_message_version` dedupes re-captured versions via `ON CONFLICT DO NOTHING` on `(account_id, change_hash)` — the docstring states it, and no test entered the conflict path: dropping the clause kept the whole suite green. Without it, every re-scan of an edited message raises IntegrityError inside the SAVEPOINT, gets swallowed by the blanket `except`, and floods one WARNING per message per cycle on a large archive — with the caller told False-as-failure instead of False-as-duplicate.

New real-engine test records the identical version twice and pins the contract: second call returns False, exactly one row exists, and nothing reaches the log (the caplog assertion is what separates DO-NOTHING from except-swallowed, since both return False). Verified by mutation: dropping the SQLite branch's `on_conflict_do_nothing` fails exactly this test.

Bead `9t6.8.27`.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage to ensure identical message versions are silently deduplicated.
  * Verified duplicate recordings return the expected result, create only one stored version, and produce no warning.

### DIFF
--- a/tests/test_db_adapter_soft_delete_upsert.py
+++ b/tests/test_db_adapter_soft_delete_upsert.py
@@ -908,3 +908,36 @@ async def test_deletion_snapshot_media_type_is_deterministic_across_rows(sqlite_
 
     # Lexicographic Media.id order: "300_62_photo" < "300_62_video".
     assert snapshot["media_type"] == "photo"
+
+
+@pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


GeiserX:
@coderabbitai full review

PR #417, opened 2026-08-26
fix(compose): use multi-arch 8.3.2 images

skywinder:
- Update the backup and viewer image pins from 8.1.0 to 8.3.2.
- Update the restricted-viewer example to use the same release.

## Summary by CodeRabbit

* **Chores**
  * Updated the Telegram backup, viewer, and example viewer containers to version 8.3.2.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   telegram-backup:
-    image: drumsergio/telegram-archive:8.1.0
+    image: drumsergio/telegram-archive:8.3.2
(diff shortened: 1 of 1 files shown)


GeiserX:
Thanks Petr, good catch. Some context on why these pins were exactly the wrong ones to be stuck on: 8.1.0 was the last amd64-only release — the workflows only started actually publishing the arm64 half of the manifest with 8.2.0 (#337). I'm also adding a CI guard so the compose pins can never drift behind a release again.

PR #449, opened 2026-09-14
fix(thumbnails): silence the harmless Pillow decompression-bomb warning

jordanfelle:
## Summary

`Image.MAX_IMAGE_PIXELS` is set to 50M as a soft pixel-count gate for thumbnail generation, but Pillow only *raises* past double that value — everything up to 100 MP decodes fine, just noisily via `DecompressionBombWarning`. JPEGs specifically are decoded via a bounded-cost draft path regardless of pixel count (`img.thumbnail()`/`draft()` scale the decode down before it happens), so a 50–100 MP JPEG hitting this warning is expected, harmless traffic — not a signal anyone acts on — but it was landing in the logs on every such photo.

## Change

- Adds `_suppress_decompression_bomb_warning()` in `src/web/thumbnails.py`, a context manager scoped to exactly `Image.DecompressionBombWarning`.
- Wraps all three `Image.open()` call sites that can hit it: the two thumbnail generators in `src/web/thumbnails.py` and the pre-generation path in `src/telegram_backup.py` (which already imports from that module).
- No change to the actual pixel-count safety logic — this only silences the known-harmless warning for images that were already being processed successfully.

## Test plan

- [x] Full existing test suite: 4182 passed, 6 skipped (PostgreSQL-dialect-only, unrelated to this change), against a real `postgres:18-alpine` service matching CI's `tests.yml` config
- [x] `ruff check .` and `ruff format --check .` — clean
- [x] Three new unit tests for the context manager: suppresses `DecompressionBombWarning`, does not suppress other warning categories, and doesn't leak suppression past the `with` block
- [x] Branch based on current upstream `main`

## Summary by CodeRabbit

- **Bug Fixes**
  - Thumbnail generation no longer emits Pillow decompression-bomb warnings during image processing.
  - Other warning types remain unaffected.


jordanfelle:
Pushed a follow-up commit addressing the CodeRabbit finding.

You're right that `warnings.catch_warnings()` around each `Image.open()` call isn't safe here: `_generate_sync`/`_generate_video_sync` are explicitly designed to run concurrently in `run_in_executor()` worker threads, and `catch_warnings()` mutates the *process-global* filter list rather than anything thread-local (context-aware warnings would fix that, but they're opt-in via `PYTHON_CONTEXT_AWARE_WARNINGS` and not this repo's default, so I'd rather not depend on every deployment setting it correctly).

Rather than asking users to set that env var, I replaced the per-call context manager with a single `warnings.filterwarnings("ignore", category=Image.DecompressionBombWarning)` registered once at module import time, before any thread starts. No context manager, no per-call global-state mutation, no race window — and no external configuration required. All three `Image.open()` call sites are back to plain `with Image.open(...) as img:`.

Tests updated accordingly (they re-import the module inside a local `catch_warnings()` block, since pytest itself wraps each test in its own `catch_warnings()` that resets the filter list). Full suite still passes (4182 passed) against a real Postgres matching CI.

GeiserX:
@coderabbitai full review

GeiserX:
Thanks for this, @jordanfelle, and for reworking it after CodeRabbit's thread-safety note.

Reviewing it turned up something that matters more than the log noise: in some cases this warning is the only log line that points at a real memory problem in thumbnail generation, so silencing it would hide exactly the thing we need to see. We're fixing the underlying limit separately and will share the details once that fix ships. Closing this one for that reason.

You were right about Pillow's thresholds, though: it only raises above twice `MAX_IMAGE_PIXELS` and warns in between.

GeiserX:
Following up now that 8.11.0 is out, as promised.

What your PR led us to: `draft()` picks a reduced decode scale and rewrites `img.size` to match. A baseline JPEG then decodes at that scale, but a progressive JPEG still allocates buffers for the full size its header declares. The viewer applied its 25 MP thumbnail cap after `draft()`, so it compared the reduced size: a large progressive JPEG passed the cap and then decoded at full size, costing hundreds of megabytes for one thumbnail. The backup's thumbnail pre-generation skipped the cap for JPEGs altogether. For such a file between 50 and 100 MP, `DecompressionBombWarning` was the only line it left in the log, which is why we kept the warning.

[8.11.0](https://github.com/GeiserX/Telegram-Archive/releases/tag/v8.11.0) (#450) checks the size the header declares against the cap before `draft()` runs, in both places. Baseline JPEGs under the cap still get the reduced decode, so your point about the draft path holds for them. The warning can still appear for files above 50 MP, since Pillow checks the size inside `Image.open()`, but it is now followed by a refused thumbnail rather than a full decode.

Thanks again for digging into Pillow's thresholds. That is what made us look closely at the draft path.

